PR #23, opened 2018-08-05
make dep sync with tf-operator to support Kubernetes 1.9

cheyang:
Hi, I am working on [arena](https://github.com/AliyunContainerService/arena) which is a CLI for data scientists, and we have already support TF-Operator. I am planing to support mpi-operator, but I noticed its Go dependency conflict to TF-Operator. Can we keep it sync to TF-operator? Thanks.

Arena is using TF-Operator as 
 library to check the pod details.  see https://github.com/AliyunContainerService/arena/blob/master/cmd/arena/commands/trainer_tensorflow.go#L266

The changes to keep the same dependency with TF-Operator in  as follow.
[k8s.io/api](https://github.com/kubeflow/tf-operator/blob/master/Gopkg.toml#L67)
[k8s.io/apimachinery](https://github.com/kubeflow/tf-operator/blob/master/Gopkg.toml#L71)
[client-go](https://github.com/kubeflow/tf-operator/blob/master/Gopkg.toml#L55)
[k8s.io/code-generator](https://github.com/kubeflow/tf-operator/blob/master/Gopkg.toml#L75)

The detail changes are:
1. Go to directory `mpi-operator`,Edit Gopkg.toml and delete Gopkg.lock， Run `dep ensure`
2. Run `./hack/update-codegen.sh`
3. Run `./hack/verify-codegen.sh`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/23)


jlewi:
@cheyang Can you update the PR to provide more info about what dependencies you are changing? It looks like you are changing the K8s version for the client?

Why is Arena dependent on the K8s version of the TFJob and MPI operators? Is arena linking in the go code so that it can be used as a library?

cheyang:
@jlewi , updated.

gaocegege:
LGTM

I also think that we should unify the vendor version.

jlewi:
Thanks.

/lgtm

@rongou @anfeng can one of you approve please?

/assign @anfeng

jlewi:
@cheyang  FWIW it doesn't seem very scalable that Arena requires compiling in client libraries for all resources that it supports and enforcing consistency across all such resources in dependencies. kubectl for example can support any CRDs in your cluster; you don't need to compile it into kubectl.

cheyang:
Thanks, it's a good point! I think the reason `kubectl` can get CRD is that https://github.com/kubernetes/client-go/blob/master/dynamic/simple.go#L231. It uses unstructured object, it's difficult to visit the spec and status info. I think I can copy the object definition into Arena  directly instead of putting them into vendor.

jlewi:
I'll merge this manually because none of the approvers are responding and tide is currently broken:
https://github.com/kubernetes/test-infra/issues/9001

rongou:
Just realized individual commits were left in. Can we squash commits and fast forward in the future? @jlewi ?

PR #24, opened 2018-08-09
copy labels to pod from podTemplate

cheyang:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/24)

### DIFF
--- a/pkg/controllers/mpi_job_controller.go
+++ b/pkg/controllers/mpi_job_controller.go
@@ -65,6 +65,8 @@ const (
 	launcherSuffix      = "-launcher"
 	workerSuffix        = "-worker"
 	gpuResourceName     = "nvidia.com/gpu"
+	labelGroupName      = "group_name"
+	labelMPIJobName     = "mpi_job_name"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/controllers/mpi_job_controller.go
```diff
@@ -485,7 +487,7 @@ func allocateGPUs(mpiJob *kubeflow.MPIJob, gpusPerNode int, done bool) (workerRe
 		if totalGPUs < gpusPerNode {
 			workerReplicas = 1
 			gpusPerWorker = totalGPUs
-		} else if totalGPUs % gpusPerNode == 0 {
+		} else if totalGPUs%gpusPerNode == 0 {
```
gaocegege: Personally, I prefer the previous way.

pkg/controllers/mpi_job_controller.go
```diff
@@ -860,11 +862,20 @@ func newLauncherRoleBinding(mpiJob *kubeflow.MPIJob) *rbacv1.RoleBinding {
 // discover the MPIJob resource that 'owns' it.
 func newWorker(mpiJob *kubeflow.MPIJob, desiredReplicas int32, gpus int) *appsv1.StatefulSet {
 	labels := map[string]string{
-		"app": mpiJob.Name + workerSuffix,
+		// "app": mpiJob.Name + workerSuffix,
```
gaocegege: If we do not need the field, we could just remove it.

pkg/controllers/mpi_job_controller.go
cheyang: It's changed by `go fmt` not me. :)

Please check https://goreportcard.com/report/github.com/kubeflow/mpi-operator

pkg/controllers/mpi_job_controller.go
cheyang: Done.

pkg/controllers/mpi_job_controller.go
gaocegege: Oh my fault, LGTM

PR #26, opened 2018-08-11
launcher also needs delete gpu requests

### LINKED ISSUE #25: Failed to launch MPI Job when specifying gpu request
make small changes of mpijob in the example,  only add gpu in requests. The yaml is as below:

```
apiVersion: kubeflow.org/v1alpha1
kind: MPIJob
metadata:
  name: mpi
spec:
  replicas: 3
  template:
    spec:
      containers:
      - image: mpioperator/tensorflow-benchmarks:latest
        name: tensorflow-benchmarks
        command: ["tail", "-f", "/dev/null"]
        resources:
          limits:
            nvidia.com/gpu: 1
          requests:
            nvidia.com/gpu: 1
```

The launcher job is not created, and checking the mpi-operator logs:

```
E0811 04:53:50.407244       1 mpi_job_controller.go:369] error syncing 'default/mpi-dist-mpijob': Job.batch "mpi-dist-mpijob-launcher" is invalid: spec.template.spec.containers[0].resources.limits: Required value: Limit must be set for non overcommitable resources
```

cheyang:
Fix #25

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/26)

### DIFF
--- a/pkg/controllers/mpi_job_controller.go
+++ b/pkg/controllers/mpi_job_controller.go
@@ -987,6 +987,9 @@ func newLauncher(mpiJob *kubeflow.MPIJob, kubectlDeliveryImage string) *batchv1.
 	if container.Resources.Limits != nil {
 		delete(container.Resources.Limits, gpuResourceName)
 	}
+	if container.Resources.Requests != nil {
+		delete(container.Resources.Requests, gpuResourceName)
+	}
 	container.VolumeMounts = append(container.VolumeMounts,
 		corev1.VolumeMount{
 			Name:      kubectlVolumeName,


cheyang:
Why setting `nvidia.com/gpu` is not allowed? In Pod spec, it's fine. The only requirement is that they should be equal.

everpeace:
@cheyang 

> The only requirement is that request and limit of nvidia.com/gpu should be equal.

ah, you're correct.  sorry, I was rushed. Then, looks good to me 👍

jlewi:
@cheyang I sent you #30 so that I can approve PRs since @rongou @anfeng don't seem to be responding.

cheyang:
Done. Thanks.

PR #28, opened 2018-08-13
Support retry times

### LINKED ISSUE #27: Need to specify the retry times of the MPI Job
Current the launcher Job's backoffLimit is 6, and it can't be modified the by end user. I suggest that the end user should be able to modify it.

And I think it can be 0 by default.

cheyang:
Fix #27

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/28)

### DIFF
--- a/pkg/controllers/mpi_job_controller.go
+++ b/pkg/controllers/mpi_job_controller.go
@@ -987,6 +987,9 @@ func newLauncher(mpiJob *kubeflow.MPIJob, kubectlDeliveryImage string) *batchv1.
 	if container.Resources.Limits != nil {
 		delete(container.Resources.Limits, gpuResourceName)
 	}
+	if container.Resources.Requests != nil {
+		delete(container.Resources.Requests, gpuResourceName)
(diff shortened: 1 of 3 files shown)


cheyang:
Thanks for your suggestions! @everpeace

pkg/controllers/mpi_job_controller.go
```diff
@@ -915,6 +915,12 @@ func newWorker(mpiJob *kubeflow.MPIJob, desiredReplicas int32, gpus int) *appsv1
 		},
 	})
 
+	// set default BackoofLimit
+	if mpiJob.Spec.BackoffLimit == nil {
+		mpiJob.Spec.BackoffLimit = new(int32)
+		*mpiJob.Spec.BackoffLimit = 0
```
rongou: Perhaps set this to 6 to be consistent with Kubernetes `Jobs`?

cheyang:
@rongou , PTAL.

PR #35, opened 2018-08-15
make job launcher no resource requirement, and prefer to put in master

### LINKED ISSUE #31: Should the limits and requests of the resource(cpu, memory) copy to launcher resource?
When I create a MPIJob which specifies cpu and memory, it means Job Launcher also requires such a mount of resources. But it's difficult to allocate such large resource for the Job launcher.

```
apiVersion: kubeflow.org/v1alpha1
kind: MPIJob
metadata:
  name: mpi
spec:
  replicas: 3
  template:
    spec:
      containers:
      - image: mpioperator/tensorflow-benchmarks:latest
        name: tensorflow-benchmarks
        command: ["tail", "-f", "/dev/null"]
        resources:
          limits:
            nvidia.com/gpu: 1
            cpu: 55
            memory: 120Gi
          requests:
            nvidia.com/gpu: 1
            cpu: 55
            memory: 120Gi
```

How do we handle such issue? The ideas in my head:

1. Remove requests and limits from Job launcher

2. Make Job launcher as worker, it also participates the computation.

@rongou @everpeace @jlewi Can you share your thoughts?

cheyang:
Fix #31

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/35)

### DIFF
--- a/pkg/apis/kubeflow/v1alpha1/types.go
+++ b/pkg/apis/kubeflow/v1alpha1/types.go
@@ -43,8 +43,13 @@ type MPIJobSpec struct {
 	// +optional
 	GPUs *int32 `json:"gpus,omitempty"`
 
+	// Run the launcher on the master.
+	// Optional: Default to false
+	// +optional
+	LauncherOnMaster bool `json:"launcherOnMaster,omitempty"`
+
 	// Optional number of retries before marking this job failed.
-	// Defaults to 0
+	// Defaults to 6
 	// +optional
 	BackoffLimit *int32 `json:"backoffLimit,omitempty"`
 
--- a/pkg/controllers/mpi_job_controller.go
+++ b/pkg/controllers/mpi_job_controller.go
@@ -87,6 +87,9 @@ const (
 	// MessageResourceSynced is the message used for an Event fired when an
 	// MPIJob is synced successfully.
 	MessageResourceSynced = "MPIJob synced successfully"
+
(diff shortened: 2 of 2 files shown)


rongou:
Hmm, not sure if it's a good idea to always run the launcher on master. Imagine a production cluster with hundreds of mpi jobs, it might adversely affect the master node. Seems like it would be better the spread the launcher pods among non-master nodes. What do you think @cheyang?

cheyang:
How about making it configurable? For example, set `RunLauncherOnMaster`. By default, it's false, means running on worker. If it's true, means running master.

The reason is that when we are running high work load training, the launcher may evicted because its QOS is BestOfEffort.

Maybe in future v1Aphla2, we can consider to separate laucher spec and worker spec.

rongou:
Are you training on GPUs? Then those nodes should have some spare CPU and memory, right? I wonder if it makes sense to change the launcher to `Guaranteed` qos.

cheyang:
Not only for GPU, we are also running some pure CPU work(like WRF), the worker will take most of the CPU and Memory in the Kubernetes Node. In this case, we don't have available resource for launcher on Kubernetes Node.

I think launcher can be Guaranteed, but the resource requests and limits is much less  than worker.

I think the root issue is a single pod spec for both launcher and worker, we may consider it in future.

cheyang:
@rongou , PTAL. Thanks.

PR #62, opened 2018-12-27
Clarify that gpusPerNode will be ignored if this is set in pod spec

terrytangyuan:
This will be ignored if the GPU resources are explicitly specified in the MPIJob pod spec.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/62)


terrytangyuan:
There seems to be an unexpected failure with Travis itself.

PR #67, opened 2019-01-07
Allow users to pass resync period for informers

terrytangyuan:
Sometimes users need to control the resync period for informers. The default is kept to be 0 (don't resync at all) in case that there are many MPIJobs for large clusters where resync might take a long time.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/67)


terrytangyuan:
Okay I am closing this for now unless there is an issue with it later.

Issue #22, opened 2018-07-20
mpi-operator do not  support schedule MPIJob with kube-arbitrator for gang schedule


GarnettWang:
Hi, @rongou, mpi-operator should support kube-arbitrator for gang schedule like tf-operator ?

launcher job and statefulset workers should as a batch job for scheduler,  they should be either both be scheduled or no one be scheduled? 

It's useful for improve gpu usage in the cluster when gpu resources are scarce in the cluster.

GarnettWang:
like tf-operator, mpi-operator add a flag `--enable-gang-scheduling` , and implement the integration with kube-arbitrator for gang schedule ?

rongou:
@GarnettWang great idea! I'm a little busy with other things at the moment, so probably won't be able to get to this for a while. Happy to review pull requests though. :)

GarnettWang:
I have a thought, this may be change the workflow of mpi-operator: 

![image](https://user-images.githubusercontent.com/23478460/43141873-9a544046-8f89-11e8-8ace-bc91e5d78680.png)

1. Create a ConfigMap that contains:
A helper shell script that can be used by mpirun in place of ssh. It invokes kubectl exec for remote execution.
A hostfile that lists the pods in the worker StatefulSet (in the form of ${job-id}-worker-0, ${job-id}-worker-1, ...), and the available slots (GPUs) in each pod.
**(ADD) A named `waitWorkersReady` shell script than will be mount into Launcher Job's init container.**

2. Create the RBAC resources (Role, ServiceAccount, RoleBinding) to allow remote execution (pods/exec). 

3. **(Add) Create PDB Object for the MPIJob Obejct.**

4. Create the worker StatefulSet that contains the desired replicas minus 1, as the launcher also does work. The command is set to sleep indefinitely. 
**(ADD) Add label or annotation `MPIJobId: $MPIJob.UID` in pod template's label field. the filed will be used for kube-batchd for gang schedule**

5. Create the launcher Job. It’s run under the ServiceAccount created in step 2. Environment variables are set to tell mpirun to use our helper script for remote execution, and the hostfile to use. The kubectl binary is delivered to an emptyDir volume through an init container.
**(ADD) add a init container named wait-workers, be responsible for waiting all statefulset workers be ready.**
**(ADD) Add label or annotation  `MPIJobId: $MPIJob.UID` in pod template's metadate.**

6. After the launcher job finishes, set the replicas to 0 in the worker StatefulSet.

[...]

PR #87, opened 2019-02-07
Support gang scheduling

terrytangyuan:
This PR adds support for gang scheduling via kube-batch. See discussion in https://github.com/kubeflow/mpi-operator/issues/22.

Currently this is through PDB but we need to switch to [PodGroup](https://github.com/kubernetes-sigs/kube-batch/blob/master/pkg/apis/scheduling/v1alpha1/types.go#L89) later (see discussion in https://github.com/kubernetes/enhancements/pull/639 for details). 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/87)


rongou:
/lgtm

Holding off approval a bit if anyone else wants to take a look.

PR #89, opened 2019-02-09
Add StartTime and CompletionTime in job status

Jeffwan:
This is not a breaking API change. I think it should be fine to add into v1alpha1.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/89)


terrytangyuan:
@Jeffwan Tests need to be updated as well since additional fields are expected.

Jeffwan:
@terrytangyuan Sure. I will add test.

Issue #92, opened 2019-02-11
MPI Operator v1alpha2 API Design Proposal


terrytangyuan:
Hi community, 

I am proposing the design for v1alpha2 API version for MPI Operator. You are very welcomed to join the discussion here if you have any questions, comments, concerns, and suggestions. Once we have a concensus from the community, we can then start working on individual items.

Here are the main API changes before we dive into the detail API spec (not including specific implementations):

* Removes deprecated fields that are GPU specific, specifically `GPUs` and `GPUsPerNode`. This is the remaining work from #75 and #85. 
* Separates `Template` into `LauncherSpec` and `WorkerSpec`. See #54 and #90.
* Replaces `MPIJobLauncherStatusType` with a more generic `MPIJobPodStatusType` that represents different states of the either launcher or worker pods.
* Adds `ReplicaStatuses` that represents statuses of all the worker replicas and removes `WorkerReplicas` since it can be inferred from `ReplicaStatuses`. See #90.

Below is the proposed API spec for v1alpha2:

```
type MPIJob struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
	Spec              MPIJobSpec   `json:"spec,omitempty"`
	Status            MPIJobStatus `json:"status,omitempty"`
}

type MPIJobList struct {
	metav1.TypeMeta `json:",inline"`
	metav1.ListMeta `json:"metadata"`
	Items           []MPIJob `json:"items"`
}

type MPIJobSpec struct {

	// Specifies the desired number of processing units the MPIJob should run on.
	// Mutually exclusive with the `Replicas` field.
	// +optional
	ProcessingUnits *int32 `json:"processingUnits,omitempty"`

	// The maximum number of processing units available per node.
	// Note that this will be ignored if the processing resources are explicitly
	// specified in the MPIJob pod spec.
	// +optional
	ProcessingUnitsPerNode *int32 `json:"processingUnitsPerNode,omitempty"`

	// The processing resource type, e.g. 'nvidia.com/gpu' or 'cpu'.
	// Defaults to 'nvidia.com/gpu'
	// +optional
	ProcessingResourceType string `json:"processingResourceType,omitempty"`

	// Specifies the number of slots per worker used in hostfile.
	// Defaults to the number of processing units per worker.
	// +optional
	SlotsPerWorker *int32 `json:"slotsPerWorker,omitempty"`

	// Run the launcher on the master.
[...]

Issue #90, opened 2019-02-09
Launcher and worker statuses do not correctly indicate the underlying states


terrytangyuan:
Launcher keeps crashing:
```
mpi-test-2-mpijob-launcher-lv2fx                                    1/1      CrashLoopBackOff             2          1m
mpi-test-2-mpijob-worker-0                                             1/1      Running                                0          1m
mpi-test-2-mpijob-worker-1                                             1/1       Running                                0          1m
```
However, from the launcher's log, one of the worker is the one that's failing and is killed (later found that it was due to OOM)
```
-------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
-------------------------------------------------------
--------------------------------------------------------------------------
mpirun.real noticed that process rank 1 with PID 39 on node mpi-test-2-mpijob-worker-1 exited on signal 9 (Killed).
--------------------------------------------------------------------------
```
Here's the description for launcher job which does not indicate any abnormal events:
```
Events:
  Type    Reason            Age   From            Message
  ----    ------            ----  ----            -------
  Normal  SuccessfulCreate  14m   job-controller  Created pod: mpi-test-mpijob-launcher-m8kw6
```

The above problems could potentially be addressed by https://github.com/kubeflow/mpi-operator/issues/12 (currently mpirun does not give us helpful error messages so maybe PMIx is a better here) and https://github.com/kubeflow/mpi-operator/issues/54 (currently only launcher pod is shown as failing but the workers are actually failing). There are other solutions too but I just wanted to link to other existing issues.

czheng94:
Any update on this issue?
I have experienced the same problem. Some workers went OOM, and the launcher failed in the end with its pod deleted by kube batch Job. 
The logs in the launcher also doesn't report any OOM error from workers.

terrytangyuan:
@czheng94 No I didn't get a chance to dig into the problem yet. This is a different issue than the log issue though.

czheng94:
@terrytangyuan Yeah this is a separate issue. 
According to my understanding, if MPI doesn't report OOM from workers to launcher (which is probably the current status), StatefulSets will just directly recreate the worker pod and launcher will restart, MPIJob will just keep failing until it exceeds 'backOffLimit'.

terrytangyuan:
@czheng94 Yes that is correct which is problematic and often waste resources. We should probably check the status regularly and report to launcher.

gaocegege:
[...]

Issue #54, opened 2018-10-09
separate out worker and launcher pod specs


rongou:
The worker pods and the launcher typically have very different resource requirements. Right now we have a single spec for them and the controller munges the launcher podspec to make it more resource efficient. We should give user more control by separating out the worker and launcher pod specs. Need to move the api version to v1alpha2.

See #53

soyuutaku:
Is there any Road-map for v1alpha2? Separate out worker and launcher is necessarily. It may add possible to add some initial containers for workers but the launcher just need to trigger the job.

terrytangyuan:
This is now fixed via `v1alpha2`.

PR #95, opened 2019-03-05
Initial v1alpha2 MPIJob API Spec

terrytangyuan:
This is the initial attempt for v1alpha2 MPIJob API Spec. See discussion in https://github.com/kubeflow/mpi-operator/issues/92 for some of the decisions. 

Some main differences from `v1alpha1` are that `v1alpha2`:

* Removes deprecated fields that are GPU specific, specifically `GPUs` and `GPUsPerNode`. This is the remaining work from #75 and #85.
* Replaces `Template` with `MPIReplicaSpecs` that uses the common `ReplicaSpec`, similar to what's done in `pytorch-operator` and `tf-operator`. See #54 and #90.
* Replaces `MPIJobLauncherStatusType` with the common `JobStatus` similar to `pytorch-operator` and `tf-operator`.

Note that `pkg/apis/kubeflow/v1alpha2/common_types.go` is copied directly from `tf-operator` since I don't believe `mpi-operator` should depend on `tf-operator`. We can switch to use a common repo once it's ready. We should continue the discussion. 

I also pushed the auto-generated v1alpha2 client code. It would be easier for you to review if you focus on changes in `pkg/apis/kubeflow/v1alpha2/types.go`. 

cc: @rongou @anfeng @jlewi @everpeace @gaocegege @Nivedita-V @madhukarkm @ywskycn @ScorpioCPH @jian-he @cheyang @richardsliu @johnugeorge @Jeffwan

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/95)


johnugeorge:
We had discussed this earlier. Do you want to keep copy of Common_types.go separately from the one used for TF/PyTorch ? i agree that the current location is not ideal and it has to kept out of the operator repos.

rongou:
I moved @terrytangyuan and @madhukarkm to approvers.

@madhukarkm, can you review this to make sure it works for our internal needs?

madhukarkm:
Sure Rong. @terrytangyuan: can you please add myself and @Nivedita-V as reviewers for this PR. 

Reg common status types, agree that refactoring to move common_types out of tf-operator would be best. Till then, seems better to have just one copy in tf-operator and dependency on it rather than two copies.

richardsliu:
I would agree that keeping one copy for common types is better than duplicating and possibly diverging the two APIs.

pkg/apis/kubeflow/v1alpha2/types.go
```diff
@@ -0,0 +1,98 @@
...
+	// the job may be active before the system tries to terminate it.
+	// Note that this takes precedence over `BackoffLimit` field.
+	// +optional
+	ActiveDeadlineSeconds *int64 `json:"activeDeadlineSeconds,omitempty"`
+
+	// Specifies the desired number of replicas the MPIJob should run on.
+	// The `PodSpec` should specify the number of processing units.
+	// Mutually exclusive with the `ProcessingUnits` fields.
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
```
richardsliu: How is this different from Replicas in https://github.com/kubeflow/tf-operator/blob/master/pkg/apis/common/v1beta2/types.go#L68?

pkg/apis/kubeflow/v1alpha2/types.go
```diff
@@ -0,0 +1,98 @@
...
+	metav1.ListMeta `json:"metadata"`
+	Items           []MPIJob `json:"items"`
+}
+
+type MPIJobSpec struct {
+
+	// Specifies the desired number of processing units the MPIJob should run on.
+	// Mutually exclusive with the `Replicas` field.
+	// +optional
+	ProcessingUnits *int32 `json:"processingUnits,omitempty"`
```
richardsliu: Consider creating a ProcessingSpec struct to contain all these fields.

pkg/apis/kubeflow/v1alpha2/types.go
terrytangyuan: Good catch. We can use the one in common.

terrytangyuan:
@madhukarkm @Nivedita-V I cannot assign you to review this PR for now since you are not member of kubeflow Github org yet. Please go ahead and review this when you get a chance. 

@johnugeorge @richardsliu @madhukarkm Agreed that we should avoid duplicating common types and possible divergence. I'll switch to use the one in `tf-operator` in a follow-up PR to keep this PR minimal for reviewing v1alpha2 MPIJob API Spec since vendor update would be very messy and hard to review.

pkg/apis/kubeflow/v1alpha2/types.go
```diff
@@ -0,0 +1,99 @@
...
+	// Specifies the desired number of processing units the MPIJob should run on.
+	// Mutually exclusive with `ReplicaSpec.Replicas` in `MPIReplicaSpecs`.
+	// +optional
+	Units *int32 `json:"units,omitempty"`
+
+	// The maximum number of processing units available per node.
+	// Note that this will be ignored if the processing resources are explicitly
+	// specified in the MPIJob pod spec.
+	// +optional
+	UnitsPerNode *int32 `json:"unitsPerNode,omitempty"`
```
Nivedita-V: Will `UnitsPerNode` be taken into consideration, if `ReplicaSpec.Replicas` is specified?

pkg/apis/kubeflow/v1alpha2/types.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// The processing resource type, e.g. 'nvidia.com/gpu' or 'cpu'.
+	// Defaults to 'nvidia.com/gpu'
+	// +optional
+	ResourceType string `json:"resourceType,omitempty"`
```
Nivedita-V: Would be good to have constants for allowed ResourceTypes

everpeace: And I would suggest using [`corev1.ResourceName`](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4162)  type instead of string.

terrytangyuan: Thanks! Done.

pkg/apis/kubeflow/v1alpha2/types.go
terrytangyuan: I think you meant `ReplicaSpec.Template` instead of `ReplicaSpec.Replicas`. I just updated the note/comment for `UnitsPerNode`.

johnugeorge:
Sounds good

terrytangyuan:
I've addressed all the comments. Please take a look again when you get a chance.

pkg/apis/kubeflow/v1alpha2/types.go
```diff
@@ -0,0 +1,111 @@
...
+	// Specifies the desired number of processing units the MPIJob should run on.
+	// Mutually exclusive with `ReplicaSpec.Replicas` in `MPIReplicaSpecs`.
+	// +optional
+	Units *int32 `json:"units,omitempty"`
+
+	// The maximum number of processing units available per node.
+	// Note that this will be ignored if the processing resources are explicitly
+	// specified in `ReplicaSpec.Template` in `MPIReplicaSpecs`.
+	// +optional
+	UnitsPerNode *int32 `json:"unitsPerNode,omitempty"`
```
madhukarkm: Are we assuming nodes in the cluster to be uniform/homogeneous?

pkg/apis/kubeflow/v1alpha2/types.go
```diff
@@ -0,0 +1,111 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []MPIJob `json:"items"`
+}
+
+type ProcessingSpec struct {
+	// Specifies the desired number of processing units the MPIJob should run on.
+	// Mutually exclusive with `ReplicaSpec.Replicas` in `MPIReplicaSpecs`.
+	// +optional
+	Units *int32 `json:"units,omitempty"`
```
madhukarkm: Will these be converted to replica pod spec.. what about memory in the case of cpu type?

pkg/apis/kubeflow/v1alpha2/types.go
```diff
@@ -0,0 +1,111 @@
...
+	// +optional
+	ResourceType ResourceType `json:"resourceType,omitempty"`
+}
+
+type MPIJobSpec struct {
+
+	// Specifies the processing spec, including number of units,
+	// units per node, resource type, etc.
+	// +optional
+	ProcessingSpec *ProcessingSpec `json:"processingSpec,omitempty"`
```
madhukarkm: Is this intended only for the worker replicas (and not for launcher replica).

pkg/apis/kubeflow/v1alpha2/common_types.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// ReplicaType represents the type of the replica. Each operator needs to define its
+// own set of ReplicaTypes.
+type ReplicaType string
+
+// ReplicaStatus represents the current observed state of the replica.
+type ReplicaStatus struct {
+	// The number of actively running pods.
+	Active int32 `json:"active,omitempty"`
```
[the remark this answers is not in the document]
terrytangyuan: Sure that should probably be added once we have a common repo. We can focus on MPIJob's spec for this PR.

pkg/apis/kubeflow/v1alpha2/common_types.go
```diff
@@ -0,0 +1,156 @@
...
+	CleanPodPolicyUndefined CleanPodPolicy = ""
+	CleanPodPolicyAll       CleanPodPolicy = "All"
+	CleanPodPolicyRunning   CleanPodPolicy = "Running"
+	CleanPodPolicyNone      CleanPodPolicy = "None"
+)
+
+// RestartPolicy describes how the replicas should be restarted.
+// Only one of the following restart policies may be specified.
+// If none of the following policies is specified, the default one
+// is RestartPolicyAlways.
```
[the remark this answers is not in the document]
terrytangyuan: Let's focus on MPIJob's spec for this PR to avoid any divergent changes in common types. I'll keep these notes unresolved so we can address them later.

pkg/apis/kubeflow/v1alpha2/types.go
terrytangyuan: Yes only for worker replicas.

pkg/apis/kubeflow/v1alpha2/types.go
terrytangyuan: Good question. This is not currently considered inside `ProcessingSpec`. We could add memory, disk, etc. to the spec. Another idea is that we could get rid of `ProcessingSpec` and make fully use of `MPIReplicaSpecs` which will allow users to provide specific pod spec. One feedback I often receive from our users is that having both `ProcessingSpec` and `MPIReplicaSpecs` are confusing. @rongou What do you think the best approach would be?

pkg/apis/kubeflow/v1alpha2/types.go
terrytangyuan: Yes, there's that assumption on this. See discussion in https://github.com/kubeflow/mpi-operator/pull/16. We should probably document this though. This is also why I think fully utilizing `MPIReplicaSpecs` would be less confusing (see my other comment).

pkg/apis/kubeflow/v1alpha2/types.go
rongou: I'm fine with getting rid of `ProcessingSpec`.

terrytangyuan: Sounds good. @madhukarkm @rongou I've just removed ProcessingSpec. Please take another look when you get a chance. This is just initial API spec. I also added some TODOs that will be addressed later once we have a common operator.

terrytangyuan:
> Sounds good to remove ProcessingSpec.. looks like ResourceType can also be removed.

@madhukarkm Thanks. Fixed.

madhukarkm:
/lgtm

Let's wait till tomorrow in case there are any other comments.

PR #100, opened 2019-03-24
 Use kube-batch as scheduler by default when gang-scheduling is enabled

jq:
Use kube-batch as scheduler by default when gang-scheduling is enabled, if scheduler is set, log warning message.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/100)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -125,32 +125,6 @@
   version = "v1.1.5"
 
 [[projects]]
-  branch = "master"
-  digest = "1:d6ec8799b1f86397187016bfc7e129e51726e9b46106b0ca52f5478f4279a4e8"
-  name = "github.com/kubeflow/mpi-operator"
-  packages = [
-    "pkg/apis/kubeflow/v1alpha1",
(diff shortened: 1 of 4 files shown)


Gopkg.lock
```diff
@@ -124,6 +124,32 @@
   revision = "1624edc4454b8682399def8740d46db5e4362ba4"
   version = "v1.1.5"
 
+[[projects]]
+  branch = "master"
+  digest = "1:d6ec8799b1f86397187016bfc7e129e51726e9b46106b0ca52f5478f4279a4e8"
+  name = "github.com/kubeflow/mpi-operator"
```
rongou: This doesn't look right. It shouldn't depend on itself.

terrytangyuan: @jq I think you need to do a `dep ensure` again and push your new changes in `Gopkg.lock` which should resolve this issue.

jq: remove it directly

rongou:
Ah I think the `solve-meta` section is a new feature in `dep`.

/lgtm
/approve

PR #101, opened 2019-03-26
Add namespace for kubectl exec cmd

fisherxu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/101)

### DIFF
--- a/pkg/controllers/mpi_job_controller.go
+++ b/pkg/controllers/mpi_job_controller.go
@@ -854,8 +854,8 @@ func newConfigMap(mpiJob *kubeflow.MPIJob, workerReplicas int, processingUnitsPe
 set -x
 POD_NAME=$1
 shift
-%s/kubectl exec ${POD_NAME} -- /bin/sh -c "$*"
-`, kubectlMountPath)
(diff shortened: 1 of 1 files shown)


pkg/controllers/mpi_job_controller.go
```diff
@@ -854,8 +854,8 @@ func newConfigMap(mpiJob *kubeflow.MPIJob, workerReplicas int, processingUnitsPe
 set -x
 POD_NAME=$1
 shift
-%s/kubectl exec ${POD_NAME} -- /bin/sh -c "$*"
-`, kubectlMountPath)
+%s/kubectl exec ${POD_NAME} -- /bin/sh -n%s -c "$*"
```
rongou: Are you sure this works? You are passing a `-n<namespace>` to `/bin/sh`, which doesn't really make sense.

fisherxu: Sorry for the delay, have addressed :) If we don't deloy the mpijob in default ns, this options is necessary.

fisherxu: In my environment, don't use the serviceaccount to connect to kube-apiserver, instead set the kubeconfig by myself. So it's not necessary in the general scenario.

PR #106, opened 2019-04-09
only create the pbdInformer when gang scheduling is enabled

jq:
only create the pbdInformer when gang scheduling is enabled

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/106)

### DIFF
--- a/cmd/mpi-operator/main.go
+++ b/cmd/mpi-operator/main.go
@@ -72,7 +72,7 @@ func main() {
 	}
 
 	var pdbInformer policyinformers.PodDisruptionBudgetInformer
-	if !enableGangScheduling {
+	if enableGangScheduling {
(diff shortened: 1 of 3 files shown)


cmd/mpi-operator/main.go
```diff
@@ -70,6 +71,10 @@ func main() {
 		kubeflowInformerFactory = informers.NewSharedInformerFactoryWithOptions(kubeflowClient, 0, informers.WithNamespace(namespace), nil)
 	}
 
+	var pdbInformer policyinformers.PodDisruptionBudgetInformer
+	if !enableGangScheduling {
```
terrytangyuan: The condition should be the other way around (same in other places)? e.g. if gang scheduling is enabled, then we initialize pdbInformer.

PR #110, opened 2019-05-29
Add leader election

fisherxu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/110)

### DIFF
--- a/cmd/mpi-operator.v1alpha2/app/server.go
+++ b/cmd/mpi-operator.v1alpha2/app/server.go
@@ -15,21 +15,34 @@
 package app
 
 import (
+	"context"
+	"fmt"
 	"os"
+	"time"
 
 	"github.com/golang/glog"
 	controllersv1alpha2 "github.com/kubeflow/mpi-operator/pkg/controllers/v1alpha2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


cmd/mpi-operator.v1alpha2/app/server.go
```diff
@@ -67,49 +93,138 @@ func Run(opt *options.ServerOption) error {
...
+		}
+	}
+
+	id, err := os.Hostname()
 	if err != nil {
-		glog.Fatalf("Error building kubeflow clientset: %s", err.Error())
+		return fmt.Errorf("failed to get hostname: %v", err)
+	}
+	// add a uniquifier so that two processes on the same host don't accidentally both become active
+	id = id + "_" + string(uuid.NewUUID())
```
fisherxu: if don't add uuid to id, when the host has same hostname, the process will all become active.

fisherxu:
/assign @rongou @terrytangyuan 

Just add K8s leader election code, and have tested, ptal.

fisherxu:
> Should do the similar change to tf and pytorch?

add uuid to id? @gaocegege

gaocegege:
@fisherxu Yeah. I think if mpi operator has the problem, then tf-operator and pytorch-operator will have the same problem, too.

PR #112, opened 2019-06-05
When MpiJob finished, delete podgroup and set worker count to 0

fisherxu:
Also change kube-batch v1alpha2 to v1alpha1, v1alpha2  is still in development.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/112)

### DIFF
--- a/pkg/controllers/v1alpha2/mpi_job_controller.go
+++ b/pkg/controllers/v1alpha2/mpi_job_controller.go
@@ -20,10 +20,10 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	podgroupv1alpha2 "github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha2"
+	podgroupv1alpha1 "github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
 	kubebatchclient "github.com/kubernetes-sigs/kube-batch/pkg/client/clientset/versioned"
-	podgroupsinformer "github.com/kubernetes-sigs/kube-batch/pkg/client/informers/externalversions/scheduling/v1alpha2"
-	podgroupslists "github.com/kubernetes-sigs/kube-batch/pkg/client/listers/scheduling/v1alpha2"
+	podgroupsinformer "github.com/kubernetes-sigs/kube-batch/pkg/client/informers/externalversions/scheduling/v1alpha1"
+	podgroupslists "github.com/kubernetes-sigs/kube-batch/pkg/client/listers/scheduling/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -308,8 +308,8 @@ func NewMPIJobController(
 		podgroupsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 			AddFunc: controller.handleObject,
 			UpdateFunc: func(old, new interface{}) {
-				newPolicy := new.(*podgroupv1alpha2.PodGroup)
-				oldPolicy := old.(*podgroupv1alpha2.PodGroup)
+				newPolicy := new.(*podgroupv1alpha1.PodGroup)
+				oldPolicy := old.(*podgroupv1alpha1.PodGroup)
 				if newPolicy.ResourceVersion == oldPolicy.ResourceVersion {
 					// Periodic re-sync will send update events for all known PodDisruptionBudgets.
 					// Two different versions of the same Job will always have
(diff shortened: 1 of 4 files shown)


pkg/controllers/v1alpha2/mpi_job_controller.go
```diff
@@ -440,12 +439,15 @@ func (c *MPIJobController) syncHandler(key string) error {
...
+	done := launcher != nil && isJobFinished(launcher)
 
-	workerSpec := mpiJob.Spec.MPIReplicaSpecs[kubeflow.MPIReplicaTypeWorker]
-	workerReplicas := *workerSpec.Replicas
+	// If MPIJob have done, workerReplicas will set to default 0.
+	var workerReplicas int32
 
 	if !done {
+		workerSpec := mpiJob.Spec.MPIReplicaSpecs[kubeflow.MPIReplicaTypeWorker]
+		workerReplicas = *workerSpec.Replicas
```
fisherxu: only set workerReplicas when launcher haven't done, If launcher have done, left  it workerReplicas as default 0

pkg/controllers/v1alpha2/mpi_job_controller.go
```diff
@@ -440,12 +439,15 @@ func (c *MPIJobController) syncHandler(key string) error {
 		return err
 	}
 	// We're done if the launcher either succeeded or failed.
-	done := launcher != nil && (launcher.Status.Succeeded == 1 || launcher.Status.Failed == 1)
+	done := launcher != nil && isJobFinished(launcher)
```
fisherxu: let's use code [#L24](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/utils.go#L24) from k8s to judge if job has finished.

pkg/controllers/v1alpha2/mpi_job_controller.go
```diff
@@ -923,16 +961,16 @@ func newLauncherRoleBinding(mpiJob *kubeflow.MPIJob) *rbacv1.RoleBinding {
 // newPodGroup creates a new PodGroup for an MPIJob
 // resource. It also sets the appropriate OwnerReferences on the resource so
 // handleObject can discover the MPIJob resource that 'owns' it.
-func newPodGroup(mpiJob *kubeflow.MPIJob, minAvailableReplicas int32) *podgroupv1alpha2.PodGroup {
-	return &podgroupv1alpha2.PodGroup{
+func newPodGroup(mpiJob *kubeflow.MPIJob, minAvailableReplicas int32) *podgroupv1alpha1.PodGroup {
+	return &podgroupv1alpha1.PodGroup{
```
gaocegege: Why do we use alpha1 now?

pkg/controllers/v1alpha2/mpi_job_controller.go
terrytangyuan: Why setting to 0? Once MPIJob is finished, all worker pods should remain undeleted there so this number should not change either, right?

fisherxu: In version v1alpha1, we just set it to 0. When the launcher has completed, What is the use of the worker :)? 
And in the product env, if don't delete this worker pod, that will be always spend the user's expense.
 How about add a field in the MPIjob, then user can decide whether to delete？

fisherxu: Like the [CleanUpPolicy](https://github.com/kubeflow/tf-operator/blob/master/pkg/apis/tensorflow/v1/types.go#L56) in tf-operator

terrytangyuan: Yes, this needs to be customized. We can replace https://github.com/kubeflow/mpi-operator/blob/master/pkg/apis/kubeflow/v1alpha2/common_types.go with https://github.com/kubeflow/common/blob/master/job_controller/api/v1/types.go, which exposes the cleanup policy.

pkg/controllers/v1alpha2/mpi_job_controller.go
fisherxu: Kube-batch still using version alpha1, alpha2 is still in development. when the alpha2 available , I can help bump :).

gaocegege: Thanks for your explanation. SGTM

pkg/controllers/v1alpha2/mpi_job_controller.go
fisherxu: The CleanPodPolicy of job will be Running/All/None, but CleanPodPolicy of statefulset will be whether to set the replica to 0... maybe can't use the common api directly
Add a commit to add a bool field CleanPodPolicy, how about it?

fisherxu: Running/All set to 0, None/"" don't set to 0?

terrytangyuan: I am not sure if it's necessary to set it to 0. We can just reuse CleanPodPolicy from kubeflow/common to decide whether to delete the pod groups.

fisherxu: Have add the CleanPodPolicy from common_type.go, now user can customize the CleanPodPolicy then decide whether delete the worker.

pkg/controllers/v1alpha2/mpi_job_controller.go
```diff
@@ -486,6 +490,13 @@ func (c *MPIJobController) syncHandler(key string) error {
 		return err
 	}
 
+	if c.enableGangScheduling && done && isCleanUpPods(mpiJob.Spec.CleanPodPolicy) {
```
fisherxu: If user set cleanupPodPolicy, will delete the podgroup alone with pods when train job have done.
@terrytangyuan  ptal :)

terrytangyuan:
@fisherxu Is it possible to use `RunPolicy` and `CleanPodPolicy` from kubeflow/common? If not, what changes are required in kubeflow/common so we can best reuse what's available to standardize the API spec.

fisherxu:
> Is it possible to use RunPolicy and CleanPodPolicy from kubeflow/common?

I think we can use the RunPolicy which contain the BackoffLimit/ActiveDeadlineSeconds/CleanPodPolicy fields used in mpi-operator now. And can we merge this first :) ? I will help change these field to RunPolicy in follow-up PRs, that should be more clear.

PR #113, opened 2019-06-10
Change the launcher done status judge logic

xieydd:
I think the launcher done status judge logic description `launcher.Status.Failed == 1`, because of `job.Status.Failed`   is the num of failed pod, and when i test, it may get nums < backoffLimit.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/113)

### DIFF
--- a/pkg/controllers/v1alpha1/mpi_job_controller.go
+++ b/pkg/controllers/v1alpha1/mpi_job_controller.go
@@ -458,7 +458,7 @@ func (c *MPIJobController) syncHandler(key string) error {
 		return err
 	}
 	// We're done if the launcher either succeeded or failed.
-	done := launcher != nil && (launcher.Status.Succeeded == 1 || launcher.Status.Failed == 1)
+	done := launcher != nil && IsJobFinished(launcher)
 
 	// TODO (terrytangyuan): Remove these flags from main.go for next major release
 	// and update deploy/*.yaml
@@ -1260,4 +1260,13 @@ func getLabelsMap(mpiJob *kubeflow.MPIJob) map[string]string {
 
 func getLauncherName(mpiJob *kubeflow.MPIJob) string {
 	return mpiJob.Name + launcherSuffix
+}
+
+func IsJobFinished(launcher *batchv1.Job) bool {
+    for _, c := range launcher.Status.Conditions {
+        if (c.Type == batchv1.JobComplete || c.Type == batchv1.JobFailed) && c.Status == corev1.ConditionTrue {
+            return true
+        }
+    }
+    return false
 }
--- a/pkg/controllers/v1alpha1/mpi_job_controller_test.go
+++ b/pkg/controllers/v1alpha1/mpi_job_controller_test.go
@@ -512,6 +512,7 @@ func TestLauncherSucceededWithGang(t *testing.T) {
 
 	launcher := newLauncher(mpiJob, "kubectl-delivery")
 	launcher.Status.Succeeded = 1
+	launcher.Status.Conditions = []batchv1.JobCondition{batchv1.JobCondition{Type: batchv1.JobComplete, Status: corev1.ConditionTrue}}
 	f.setUpLauncher(launcher)
 
 	mpiJobCopy := mpiJob.DeepCopy()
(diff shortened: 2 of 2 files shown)


pkg/controllers/v1alpha1/mpi_job_controller.go
```diff
@@ -1261,3 +1261,16 @@ func getLabelsMap(mpiJob *kubeflow.MPIJob) map[string]string {
 func getLauncherName(mpiJob *kubeflow.MPIJob) string {
 	return mpiJob.Name + launcherSuffix
 }
+
+func IsJobFinished(launcher *batchv1.Job) bool {
+    if launcher == nil {
```
fisherxu: Thanks for your fix:) We have the judge `launcher != nil &&` before, So don't need this line.

xieydd: done

terrytangyuan:
Thanks. Could you fix the build?

xieydd:
> Thanks. Could you fix the build?

I wil fix the build, but because of the gfw, maybe a little later. :(

fisherxu:
Could you squash the commits? :)

pkg/controllers/v1alpha1/mpi_job_controller_test.go
```diff
@@ -549,8 +551,10 @@ func TestLauncherFailed(t *testing.T) {
 	mpiJob := newMPIJob("test", int32Ptr(64), &startTime, nil)
 	f.setUpMPIJob(mpiJob)
 
+	mpiJob.Status.LauncherStatus = kubeflow.LauncherFailed
```
fisherxu: what's the usage of this line? mpiJobCopy has the same operation

pkg/controllers/v1alpha1/mpi_job_controller_test.go
```diff
@@ -549,8 +551,10 @@ func TestLauncherFailed(t *testing.T) {
...
 	launcher := newLauncher(mpiJob, "kubectl-delivery")
 	launcher.Status.Failed = 1
+	launcher.Status.Conditions = []batchv1.JobCondition{batchv1.JobCondition{Type: batchv1.JobComplete, Status: corev1.ConditionTrue}}
```
fisherxu: Condition should be `Failed` here?

xieydd: if i set Failed, it will not pass ci test. Prev logic is `done = launcher != nil && mpiJob.Status.Failed = 1 || mpiJob.Status.Successed = 1`, prev set `launcher.Status.Failed = 1`. When the launcher is not null, done will be true. So i just to make sure the done is true.

pkg/controllers/v1alpha1/mpi_job_controller_test.go
xieydd: I will delete the line , my bad.

xieydd:
> Could you squash the commits? :)

If there is no problem, i will squash the commits.

pkg/controllers/v1alpha1/mpi_job_controller_test.go
fisherxu: A little confuse what's the difference of Failed and Completed here? the new isJobFinished func we cover all this condition

xieydd: `job.Status.Failed`  is the num of failed pod,  it may get nums < backoffLimit. When launcher is not null and job.Status.Failed = 1 , `done` will be true. But in real world, it will restart laucher because the `backofflimit=6 `default .

terrytangyuan:
/lgtm
Please ping me when ready to merge.

xieydd:
ping @terrytangyuan ,I squash the commits.

terrytangyuan:
Actually @fisherxu, could you take a look to see if the changes addressed your comments?

pkg/controllers/v1alpha1/mpi_job_controller_test.go
fisherxu: Yes, I know this case, and in v1alpha2 I also have change to isJobFinished. Here my question is the test func is `TestLauncherFailed`, but we set the condition in L557 to JobComplete which mean success, not the Failed?

xieydd: Yep, you are right. I should set it be Failed and delete `launcher.Status.Failed = 1`. :)

fisherxu: Yes, this make sense to me :)  If you are interesting with this, suggest you to try v1alpha2, the api is more clear :)

fisherxu:
/lgtm

@terrytangyuan  Let's get this merged :)

pkg/controllers/v1alpha1/mpi_job_controller_test.go
xieydd: You can change the v1alphav2 test file. @fisherxu

fisherxu: Yeah sure.

PR #118, opened 2019-06-13
add more CI check

wackxu:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/118)


wackxu:
@rongou @terrytangyuan @gaocegege Add more CI check and fix the failed UT, PTAL

Gopkg.toml
```diff
@@ -19,3 +25,16 @@ required = ["k8s.io/code-generator/cmd/client-gen"]
 [[constraint]]
   name = "github.com/stretchr/testify"
   version = "1.3.0"
+
+[[constraint]]
+  name = "github.com/kubernetes-sigs/kube-batch"
+  branch = "master"
```
wackxu: why we use master branch for kube-batch, because for v0.3 and v0.4 release, the generate code is not updated and the UT will fail because of the generate code of kube-batch. I took a look at that and found that there miss a commit in the v0.3 and v0.4 release about update generate code. So we use the master branch.

.gitignore
```diff
@@ -1,3 +1,2 @@
 .idea/
 .vscode/
-vendor/
```
rongou: Is there a reason to check in the vendor tree?

wackxu: The main reason is that it is not convenient for some developers to install the vendor due to some network reasons and in other operators they always has the vendor,  so I think we should add the vendor to the code tree.

xieydd: LGTM for vender tree.

rongou:
I'd prefer we don't check in the vendor directory, but I don't feel too strongly about it. @terrytangyuan you have any preference?

terrytangyuan:
I am fine with the vendor tree but that can probably be done with a separate PR. Otherwise this PR is very hard to review.

wackxu:
> I am fine with the vendor tree but that can probably be done separately. Otherwise this PR is very hard to review. @wackxu Could you revert vendor tree changes in this PR?

@terrytangyuan I have a single commit for add vendor and you can just review other three pr.

wackxu:
https://github.com/kubeflow/mpi-operator/pull/120 has been merged and I do not rebase the upstream master in this pr,so I will fix the failed ut in another pr

PR #119, opened 2019-06-13
init kubebatchClient before use

wackxu:
/assign @gaocegege @terrytangyuan

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/119)


gaocegege:
How can we run the code without the PR before?

wackxu:
> How can we run the code without the PR before?

the v1alpha2 version is in developing and is not fully tested and there are some work need to do to make v1alpha2 version stable.

PR #129, opened 2019-07-03
fix: Remove the hard coded kind and check version

gaocegege:
Since we do not support conversion webhook now, I think we do not deal with v1alpha1 resources in this operator.

Thus I think we could check group, version and kind in handleobject before we run the sync logic.

/assign @wackxu @terrytangyuan 

Signed-off-by: Ce Gao <gaoce@caicloud.io>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/129)


gaocegege:
Please take a look, thanks for your time :smile:

PR #136, opened 2019-07-23
Single_deploy

vtlrazin:
Single YAML deployment file with CRD, namespace, rbac, and mpi-operator sections.
This will simplify deployment for ANSIBLE-based installations as well as for other installations.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/136)


vtlrazin:
I signed it!

deploy/mpi-operator.yaml
```diff
@@ -0,0 +1,194 @@
...
+#      MPI-operator container deployment set
+# --------------------------------------------------
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: mpijobs.kubeflow.org
+spec:
+  group: kubeflow.org
+  version: v1alpha2
```
terrytangyuan: This is only the v1alpha2 CRD. Perhaps we should only combine the yaml files besides the CRDs so we don't need to keep two files with a lot of overlapped content, e.g RBAC, etc.

vtlrazin: CRD with v1alpha1 is supported by mpioperator/mpi-operator:latest or customer must use mpioperator/mpi-operator:v1alpha1 version?

terrytangyuan: Sorry I somehow missed this question. No, v1alpha1 would not work with the latest images.

vtlrazin: So, we can leave only the v2 version on the master branch. But single yaml deployment file will make deployment easier.

vtlrazin: Do you know how to resolve LGTM label issue?

terrytangyuan: cc @rongou to double check whether we need to make v1 version of it and whether broken existing links are fine.

rongou: Yeah I think it's fine to ignore v1.

vtlrazin: Hi,
How will we continue? Do I need to do something?

terrytangyuan: Thank you! Approving now.

PR #141, opened 2019-08-21
add job priority for kube-batch scheduling

4everming:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/141)

### DIFF
--- a/deploy/crd/crd-v1alpha2.yaml
+++ b/deploy/crd/crd-v1alpha2.yaml
@@ -23,6 +23,8 @@ spec:
             slotsPerWorker:
               type: integer
               minimum: 1
+            jobPriority:
+              type: string
             mpiReplicaSpecs:
               properties:
                 Launcher:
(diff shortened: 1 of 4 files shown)


4everming:
@googlebot I signed it!

xieydd:
@4everming How about fixing the CI error.

4everming:
> @4everming How about fixing the CI error.

Got it.

deploy/crd/crd-v1alpha2.yaml
```diff
@@ -23,6 +23,8 @@ spec:
             slotsPerWorker:
               type: integer
               minimum: 1
+            jobPriority:
+              type: string
```
rongou: Why is this a string? Shouldn't it be an integer?

4everming: Hi rongou, jobPriority refers to PriorityClassName just like "system-node-critical" or any other priority.  That is consistent with PriorityClassName in PodGroupSpec of kube-batch scheduling.

terrytangyuan:
@4everming Should this be done in kubeflow/common instead so that If there are more parameters like this in the future for scheduling, we don’t have to keep adding fields to MPIJobSpec. Related discussion: https://github.com/kubeflow/common/issues/11

terrytangyuan:
Closing this since the changes have been submitted to kubeflow/common.

PR #142, opened 2019-09-03
Add default init container ResourceRequirements

czheng94:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/142)

### DIFF
--- a/pkg/controllers/v1alpha2/mpi_job_controller.go
+++ b/pkg/controllers/v1alpha2/mpi_job_controller.go
@@ -17,9 +17,10 @@ package v1alpha2
 import (
 	"bytes"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"reflect"
 	"time"
+
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/golang/glog"
 	podgroupv1alpha1 "github.com/kubernetes-sigs/kube-batch/pkg/apis/scheduling/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/controllers/v1alpha2/mpi_job_controller.go
```diff
@@ -1198,6 +1201,16 @@ func (c *MPIJobController) newLauncher(mpiJob *kubeflow.MPIJob, kubectlDeliveryI
 				MountPath: configMountPath,
 			},
 		},
+		Resources: corev1.ResourceRequirements{
```
terrytangyuan: Should we expose this to `MPIJobSpec` similar to `MPIReplicaSpecs`? What if user needs resources other than cpu and memory?

yuzisun: This is only for init container though, not sure if user needs to customize it.

terrytangyuan: What if their requirements are different, e.g. larger than these default values?

czheng94: In what cases would user want to customize the init container?

terrytangyuan: @gaocegege @rongou Any thoughts on this?

rongou: Seems fine. The init container really just copies the `kubectl` binary and does some initial checking. Resource usage should be minimal.

PR #186, opened 2020-02-10
Use operator's namespace for initializing informers and endpoints/leaderElection

terrytangyuan:
Right now `namespace` variable is the Kubeflow namespace passed from environment variable. We should re-assign it to be operator-level namespace so that endpoints meta and leader election could work correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/186)

### DIFF
--- a/cmd/mpi-operator.v1alpha2/app/server.go
+++ b/cmd/mpi-operator.v1alpha2/app/server.go
@@ -79,11 +79,14 @@ func Run(opt *options.ServerOption) error {
 		glog.Infof("%s not set, use default namespace", v1alpha2.EnvKubeflowNamespace)
 		namespace = metav1.NamespaceDefault
 	}
+	if opt.Namespace != "" {
+		namespace = opt.Namespace
+	}
 
-	if opt.Namespace == corev1.NamespaceAll {
+	if namespace == corev1.NamespaceAll {
 		glog.Info("Using cluster scoped operator")
 	} else {
-		glog.Infof("Scoping operator to namespace %s", opt.Namespace)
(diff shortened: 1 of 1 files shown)


junhuihuang:
___
*[cmd/mpi-operator.v1alpha2/app/server.go, line 82 at r1](https://reviewable.io/reviews/kubeflow/mpi-operator/186#-M1PHcuN99o8vh32M0Zt:-M1PHOjF5F8izF0IyP1Z:b-nr18bi) ([raw file](https://github.com/kubeflow/mpi-operator/blob/2da82acca310cad80697097809bc4850c8f32661/cmd/mpi-operator.v1alpha2/app/server.go#L82)):*
> ```Go
> 		namespace = metav1.NamespaceDefault
> 	}
> 	if opt.Namespace != "" {
> ```

Bug: `opt.Namespace` default value is `corev1.NamespaceAll`, and `corev1.NamespaceAll = ""`, so `namespace` can never be `corev1.NamespaceAll` and user cannot use `NamespaceAll` to monitor

terrytangyuan:
I don’t think I am following. Could you send a PR to fix it and illustrate what the right logic should be?

junhuihuang:
Thanks your time.

After merged this PR, I cannot using **cluster scoped** mpi-operator. Motoring mpijobs in all namespaces, not only the some namespace.

Resetting this PR can fix my problem.

terrytangyuan:
@junhuihuang Do you mind creating an issue on GitHub to track with reproducible example if possible and error log?

PR #188, opened 2020-02-12
add scheduling groups of kube-batch/volcano for role

carmark:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/188)

### DIFF
--- a/deploy/mpi-operator.yaml
+++ b/deploy/mpi-operator.yaml
@@ -145,6 +145,14 @@ rules:
   - mpijobs/status
   verbs:
   - "*"
+- apiGroups:
+  - scheduling.incubator.k8s.io
+  - scheduling.sigs.dev
+  resources:
+  - queues
+  - podgroups
+  verbs:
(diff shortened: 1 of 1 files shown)


deploy/mpi-operator.yaml
```diff
@@ -145,6 +145,14 @@ rules:
   - mpijobs/status
   verbs:
   - "*"
+- apiGroups:
+  - scheduling.incubator.k8s.io
```
terrytangyuan: Thanks for the PR. Could you explain why this is needed?

gaocegege: I think it is to give the controller access to volcano's CRDs such as Queue and PodGroup to support running mpi-operator with volcano

terrytangyuan: Okay sounds good. Are there any code changes required in order to support that?

carmark: none

PR #192, opened 2020-02-18
set NVIDIA_VISIBLE_DEVICES to none

carmark:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/mpi-operator/192)


pkg/controllers/v1alpha2/mpi_job_controller.go
```diff
@@ -1235,7 +1235,7 @@ func (c *MPIJobController) newLauncher(mpiJob *kubeflow.MPIJob, kubectlDeliveryI
 		// issues with scheduler/container technologies.
 		corev1.EnvVar{
 			Name:  "NVIDIA_VISIBLE_DEVICES",
-			Value: "",
+			Value: "none",
```
terrytangyuan: Seems like this is different than “”. See https://github.com/NVIDIA/nvidia-container-runtime/blob/master/README.md. Could you double check?

carmark: Yes, you are right.

PR #215, opened 2020-04-13
add monitor metrics

runzhliu:
Fully based on [tf-operator](https://github.com/kubeflow/tf-operator/pull/1018), add the metrics monitor by Prometheus in mpi-operator, which can report the Mpijob metrics like creation, deletion, completion and failure.

### DIFF
--- a/cmd/mpi-operator.v1alpha2/app/options/options.go
+++ b/cmd/mpi-operator.v1alpha2/app/options/options.go
@@ -27,6 +27,7 @@ type ServerOption struct {
 	MasterURL            string
 	KubectlDeliveryImage string
 	Threadiness          int
+	MonitoringPort       int
 	PrintVersion         bool
 	GangSchedulingName   string
 	Namespace            string
@@ -59,6 +60,9 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 		`How many threads to process the main logic`)
 
 	fs.BoolVar(&s.PrintVersion, "version", false, "Show version and quit")
+
+	fs.IntVar(&s.MonitoringPort, "monitoring-port", 8443,
+		`Endpoint port for displaying monitoring metrics. It can be set to "0" to disable the metrics serving.`)
 
 	fs.StringVar(&s.GangSchedulingName, "gang-scheduling", "", "Set gang scheduler name if enable gang scheduling.")
 
--- a/cmd/mpi-operator.v1alpha2/app/options/options.go
+++ b/cmd/mpi-operator.v1alpha2/app/options/options.go
@@ -61,7 +61,7 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 
 	fs.BoolVar(&s.PrintVersion, "version", false, "Show version and quit")
 
-	fs.IntVar(&s.MonitoringPort, "monitoring-port", 8443,
+	fs.IntVar(&s.MonitoringPort, "monitoring-port", 0,
 		`Endpoint port for displaying monitoring metrics. It can be set to "0" to disable the metrics serving.`)
 
 	fs.StringVar(&s.GangSchedulingName, "gang-scheduling", "", "Set gang scheduler name if enable gang scheduling.")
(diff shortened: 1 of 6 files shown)


cmd/mpi-operator.v1alpha2/app/server.go
```diff
@@ -17,6 +17,8 @@ package app
 import (
 	"context"
 	"fmt"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
```
carmark, line 21: should you please move these two packages to second part?

cmd/mpi-operator.v1alpha2/main.go
```diff
@@ -16,19 +16,38 @@ package main
...
 	"github.com/golang/glog"
 
 	"github.com/kubeflow/mpi-operator/cmd/mpi-operator.v1alpha2/app"
 	"github.com/kubeflow/mpi-operator/cmd/mpi-operator.v1alpha2/app/options"
 )
 
+func startMonitoring(monitoringPort int) {
+	if monitoringPort != 0 {
+		go func() {
+			glog.Infof("Setting up client for monitoring on port: %s", strconv.Itoa(monitoringPort))
```
carmark, line 33:
```
glog.Infof("Setting up client for monitoring on port: %d", monitoringPort)
```

what about the above one?

cmd/mpi-operator.v1alpha2/main.go
```diff
@@ -16,19 +16,38 @@ package main
...
+			http.Handle("/metrics", promhttp.Handler())
+			err := http.ListenAndServe(fmt.Sprintf(":%s", strconv.Itoa(monitoringPort)), nil)
```
carmark, line 35: same as above one.

carmark:
If you update the vendor code, could you please separate this commit into two parts?

cmd/mpi-operator.v1alpha2/app/options/options.go
```diff
@@ -60,6 +61,9 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 
 	fs.BoolVar(&s.PrintVersion, "version", false, "Show version and quit")
 
+	fs.IntVar(&s.MonitoringPort, "monitoring-port", 8443,
```
terrytangyuan, line 64: Should this be disabled by default in case prometheus is not available?

carmark: I think that's all right, prometheus will fetch data from this port. But the default port of monitoring should be 8080?

terrytangyuan: This implementation returns `glog.Error` if the port is not set up for Prometheus: https://github.com/kubeflow/mpi-operator/pull/215/files#diff-58e97bee9d8731245689342bcce48a7dR37

cmd/mpi-operator.v1alpha2/app/server.go
runzhliu: Sure ;)

runzhliu:
> If you update the vendor code, could you please separate this commit into two parts?

all right.

terrytangyuan:
Actually I am removing the vendor directories in https://github.com/kubeflow/mpi-operator/pull/218. Please revert the changes related to vendor and then rebase. @runzhliu @carmark

runzhliu:
> Actually I am removing the vendor directories in #218. Please revert the changes related to vendor and then rebase. @runzhliu @carmark

Good move!

cmd/mpi-operator.v1alpha2/app/options/options.go
terrytangyuan: @runzhliu See my comment above on `glog.Error`

runzhliu: @terrytangyuan I have set the default value 0, so the metrics monitor wouldn't be triggered on.

Issue #217, opened 2020-04-13
Graduate MPI Operator to v1


terrytangyuan:
This issue tracks the progress of graduating MPI Operator to v1. Please check out the [application requirements](https://github.com/kubeflow/community/blob/master/guidelines/application_requirements.md) for details. Each item in the table below includes a check box that marks the completion of that particular requirement.

If you are interested in working on any of the items below, please leave a comment on this issue.

### Configuration and deployment

| Description | Category | Explanation |
|-------------|----------|-------------|
| <ul><li>- [x] Kustomize package</li></ul>  | Required  | <ul><li>Kubeflow has standardized on Kustomize for configuring and deploying packages </ul> |
| <ul><li>- [x] Application CR</li></ul>  | Required  | <ul><li>Kubeflow has standardized on using the Kubernetes Application CR to provide consistent metadata and monitoring about applications <li> Application CR should be an owner of all resources so that deleting the application CR uninstalls the application </ul> |
| <ul><li>- [x] ```app.kubernetes.io``` labels on every resource</li></ul>  | Required  | <ul><li>Every resource that is part of the application should include the [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) recommended by Kubernetes, currently: <ul><li>`app.kubernetes.io/name` <li>`app.kubernetes.io/instance` <li>`app.kubernetes.io/version` <li>`app.kubernetes.io/component` <li>`app.kubernetes.io/part-of` <li>`app.kubernetes.io/managed-by`</ul> <li>See example [here](https://github.com/kubeflow/manifests/blob/v0.6.1/tf-training/tf-job-operator/overlays/application/application.yaml#L8-L13) </ul> |
| <ul><li>- [x] Images listed in kustomization.yaml</li></ul>  | Required  | <ul><li>All docker images used by the application should be listed in one or more kustomization.yaml files <li>This facilitates mirroring the images to different docker registries </ul> |
| <ul><li>- [x] Upgradeability</li></ul> | Required  | <ul><li>Applications must support upgrading between consecutive major and minor releases </ul> |
| <ul><li>- [ ] Separate cluster scoped and namespace scoped resources</li></ul> | Recommended  | <ul><li>To the extent possible cluster scoped resources should be installable separately (e.g. via a separate kustomize package) <li> This allows cluster admins to install only the cluster scoped resources <li> Clear documentation of cluster and namespace scoped behavior </ul> |
[...]

PR #221, opened 2020-04-15
Remove BackoffLimit and ActiveDeadlineSeconds from MPIJobSpec

terrytangyuan:
In favor of kubeflow/common's `RunPolicy`. Part of #217.


carmark:
I do like this PR, ;-)

carmark:
Remove these two fields broke the API compatibility, should we move it to new API version?

terrytangyuan:
I just added v1 API and controllers in https://github.com/kubeflow/mpi-operator/pull/225. I will re-work this PR to modify v1 API instead.

terrytangyuan:
@carmark I've updated this PR to modify v1 API instead. Please take another look.

PR #237, opened 2020-04-22
rename hovorod to horovod

malixian:
The directory name should be horovod instead of hovorod


carmark:
@malixian could you please sign the cla?

malixian:
> Thanks! Could you remove “-cpu” in the folder name so this folder can be used for future GPU examples as well.

all right

PR #248, opened 2020-05-10
Fix the wrong build path in Dockerfile

XinkaiWang515:
Fix the wrong build image copy path in Dockerfile.


XinkaiWang515:
@googlebot I fixed it.

PR #251, opened 2020-05-11
fix dockerfile copy path

rongou:
@terrytangyuan

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /go/src/github.com/kubeflow/mpi-operator
 RUN make
 
 FROM gcr.io/distroless/base-debian10:latest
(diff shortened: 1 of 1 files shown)


NikeNano:
@rongou how are the rebuild of the image triggered? Can’t see any new images and don’t understand if there is any auto builds to dockerhub.

XinkaiWang515:
@rongou Hi, Rongou. Is the  '_output' a hidden directory? Can you explain why does this change work? Thank you so much.

rongou:
@NikeNano it should build on dockerhub, but it might take a while.

@XinkaiWang515 the `Makefile` puts the output in `_output/cmd/bin` (see https://github.com/kubeflow/mpi-operator/blob/master/Makefile#L1).

XinkaiWang515:
@rongou I see. Thanks for your help!

PR #264, opened 2020-06-15
Add nodeSelector and tolerations in Helm Charts

runzhliu:
I think it would be better if we could set the `nodeSelector` and `tolerations` while installing the mpi-operator :)


terrytangyuan:
We need to upgrade helm charts to use v1 controller. Do you want to help that out as well? :-)

runzhliu:
@terrytangyuan Sure, I am working on it :)

PR #282, opened 2020-07-24
[WIP] add support for horovod elastic

zw0610:
This pr contains the following changes for mpi-operator v1:

1. deliver `hostfile` to both launcher and workers
2. update `hostfile` (via configMap) after its creation and reflect the working pods only
3. deliver `discover_hosts.sh` under `/etc/mpi` with script mode

Here remains two questions for support horovod elastic:
1. do we add support for changing worker replicas number after the mpi-job is created?
2. shall we change the function name of `getOrCreateConfigMap` to `getOrCreateOrUpdateConfigMap`? :)

-----

Since it's my first pr to kubeflow, please do let me know anything I can improve.

### DIFF
--- a/pkg/controllers/v1/mpi_job_controller.go
+++ b/pkg/controllers/v1/mpi_job_controller.go
@@ -62,6 +62,7 @@ const (
 	configMountPath         = "/etc/mpi"
 	kubexecScriptName       = "kubexec.sh"
 	hostfileName            = "hostfile"
+	discoverHostsScriptName = "discover_hosts.sh"
 	kubectlDeliveryName     = "kubectl-delivery"
 	kubectlTargetDirEnv     = "TARGET_DIR"
 	kubectlVolumeName       = "mpi-job-kubectl"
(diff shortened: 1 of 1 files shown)


zw0610:
@googlebot I signed it!

terrytangyuan:
@zw0610 Thanks! Your commit does not seem to be using the correct username or email.

pkg/controllers/v1/mpi_job_controller.go
```diff
@@ -62,6 +62,7 @@ const (
 	configMountPath         = "/etc/mpi"
 	kubexecScriptName       = "kubexec.sh"
 	hostfileName            = "hostfile"
+	discoverHostsScriptName = "discover_hosts.sh"
```
terrytangyuan, line 65: This assumes that the image contains this script, right?

pkg/controllers/v1/mpi_job_controller.go
```diff
@@ -1039,10 +1058,20 @@ shift
...
+			buffer.WriteString(fmt.Sprintf("%s slots=%d\n", name, slots))
+		}
 	}
 	for i := 0; i < int(workerReplicas); i++ {
-		buffer.WriteString(fmt.Sprintf("%s%s-%d slots=%d\n", mpiJob.Name, workerSuffix, i, slots))
+		name := fmt.Sprintf("%s%s-%d", mpiJob.Name, workerSuffix, i)
+		if runningPods == nil {
+			buffer.WriteString(fmt.Sprintf("%s slots=%d\n", name, slots))
+		} else if _, found := runningPods[name]; found {
+			buffer.WriteString(fmt.Sprintf("%s slots=%d\n", name, slots))
```
terrytangyuan, line 1073: This can be refactored

terrytangyuan:
> * do we add support for changing worker replicas number after the mpi-job is created?

This is not supported currently but we should support that.

> * shall we change the function name of `getOrCreateConfigMap` to `getOrCreateOrUpdateConfigMap`? :

That sounds strange. Perhaps split into two functions `getOrCreateConfigMap` and `updateConfigMap`?

zw0610:
@googlebot I fixed it.

pkg/controllers/v1/mpi_job_controller.go
zw0610: Right, mpi-operator can provide a *default* `discover_hosts.sh` to users and still open to `discover_hosts.sh` alternatives.

zw0610:
I believe this elastic approach (modifying `hostfile` immediately once the pod status is changed from running) is not compatible with the tests. And it could also be risky for the initial launching stage where some pods are just not ready.

An alternative will be add a second copy of `hostfile`, which reflects the pods status, and keep the original `hostfile` as it used to be.

terrytangyuan:
Any update on this?

PR #303, opened 2020-12-15
Cleans up the unused go.sum dependencies

naveensrinivasan:
Cleans up the unused `go.sum` dependencies


terrytangyuan (commented):
Could you add a check in the CI that checks the tidiness as well?

naveensrinivasan:
> Could you add a check in the CI that checks the tidiness as well?

Can you point me to the CI? Probably that can be a separate PR.

terrytangyuan:
The CI steps are defined here: https://github.com/kubeflow/mpi-operator/blob/master/.travis.yml

naveensrinivasan:
@terrytangyuan I would suggest that be a separate PR.

terrytangyuan (approved):
Sure. We probably don't want to use Travis CI anyways and we need to switch to use GitHub Actions. 

/lgtm
/approve

naveensrinivasan:
@terrytangyuan sounds good. I was thinking the same as to use GitHub actions compared to Travis.

PR #317, opened 2021-01-14
update horovod version for the tensorflow benchmark example

rongou:
Looks like people are running the tensorflow benchmark example as is, probably should update the version.

@terrytangyuan


terrytangyuan (approved):
/lgtm
/approve

PR #330, opened 2021-02-26
Remove launcherRunsWorkload flag

### LINKED ISSUE #329: [feature] Remove launcherRunsWorkload flag
Now we have a flag to determine that if we should use GPU in the launcher. 

But it is not expected in our situation. If we mount a GPU card to the launcher but the flag is disabled, we cannot see this card because we will set NVIDIA_VISIBLE_DEVICES to empty and libnvidia-container will not mount drivers to the pod.

We can remove the flag and rely on the launcher.resources.gpu to determine it.

Garrybest:
fix #329


Garrybest:
@googlebot I signed it!

PR #331, opened 2021-02-26
Remove launcherRunsWorkload flag

### LINKED ISSUE #329: [feature] Remove launcherRunsWorkload flag

Garrybest:
fix #329

### DIFF
--- a/pkg/controllers/v1/mpi_job_controller.go
+++ b/pkg/controllers/v1/mpi_job_controller.go
@@ -157,9 +157,6 @@ type MPIJobController struct {
 
 	// To allow injection of updateStatus for testing.
 	updateStatusHandler func(mpijob *kubeflow.MPIJob) error
-
-	// To allow launcher run workerload when launcher pod has GPU.
-	launcherRunsWorkload bool
 }
 
 // NewMPIJobController returns a new MPIJob controller.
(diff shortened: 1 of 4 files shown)


pkg/controllers/v1/mpi_job_controller.go
```diff
@@ -1476,6 +1471,7 @@ func isCleanUpPods(cleanPodPolicy *common.CleanPodPolicy) bool {
 	return false
 }
 
+// isGPULauncher check whether launcher needs GPU.
```
gaocegege, line 1474:
```suggestion
// isGPULauncher checks whether the launcher needs GPU.
```

gaocegege (commented):
Generally LGTM

gaocegege:
Please sign the CLA

Garrybest:
@googlebot I fixed it.

Garrybest:
@googlebot I fixed it.

pkg/controllers/v1/mpi_job_controller.go
Garrybest: Thanks for reviewing. 😄

terrytangyuan (commented):
/cc @allenzfan @carmark

xieydd (approved):

carmark (approved):

terrytangyuan (approved):

PR #332, opened 2021-02-28
add support for elastic horovod on v1

zw0610:
@gaocegege 

Here is the prototyping of adding support for Elastic Horovod. It mainly consists of three sections of modification:

- [x] Remove Pods when Worker Replicas is scaled down
- [x] Add `discover_hosts.sh` to the ConfigMap and mount it under `/etc/mpi/`
- [x] Updating the ConfigMap when the content of `discover_hosts.sh` changes

Since it is only the first try, there remains multiple places where we can improve, such like the the way the controller determines whether the content of `discover_hosts.sh` changed. These improvements will be included in this pr later.

### DIFF
--- a/examples/horovod/tensorflow-mnist-elastic.yaml
+++ b/examples/horovod/tensorflow-mnist-elastic.yaml
@@ -1,0 +1,39 @@
+apiVersion: kubeflow.org/v1
+kind: MPIJob
+metadata:
+  name: tensorflow-mnist
+spec:
+  slotsPerWorker: 1
+  cleanPodPolicy: Running
+  mpiReplicaSpecs:
+    Launcher:
+      replicas: 1
+      template:
+        spec:
+          containers:
+          - image: docker.io/kubeflow/mpi-horovod-mnist
+            name: mpi-launcher
+            command:
+            - horovodrun
+            args:
+            - -np
+            - "2"
+            - --host-discovery-script
+            - /etc/mpi/discover_hosts.sh
+            - python
+            - /examples/tensorflow_mnist.py
+            resources:
+              limits:
+                cpu: 1
+                memory: 2Gi
+    Worker:
+      replicas: 2
+      template:
+        spec:
+          containers:
+          - image: docker.io/kubeflow/mpi-horovod-mnist
+            name: mpi-worker
+            resources:
+              limits:
+                cpu: 2
+                memory: 4Gi
--- a/pkg/controllers/v1/mpi_job_controller.go
+++ b/pkg/controllers/v1/mpi_job_controller.go
@@ -18,6 +18,8 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -29,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -62,6 +65,7 @@ const (
 	configMountPath         = "/etc/mpi"
 	kubexecScriptName       = "kubexec.sh"
 	hostfileName            = "hostfile"
+	discoverHostsScriptName = "discover_hosts.sh"
 	kubectlDeliveryName     = "kubectl-delivery"
 	kubectlTargetDirEnv     = "TARGET_DIR"
 	kubectlVolumeName       = "mpi-job-kubectl"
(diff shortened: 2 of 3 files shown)


pkg/apis/kubeflow/v1/types.go
```diff
@@ -37,6 +37,14 @@ type MPIJobList struct {
 	Items           []MPIJob `json:"items"`
 }
 
+// ElasticPolicy describes which Elastic Mode or none is enabled.
+type ElasticPolicy string
+
+const (
+	ElasticPolicyUndefined ElasticPolicy = ""
+	ElasticPolicyHorovod ElasticPolicy = "ElasticHorovod"
```
terrytangyuan, line 45: Can we detect this automatically?

terrytangyuan:
Would you like to close https://github.com/kubeflow/mpi-operator/pull/282 in favor of this PR then?

zw0610:
> Would you like to close #282 in favor of this PR then?

good idea. I shall include an example later.

pkg/apis/kubeflow/v1/types.go
zw0610: I think not. The only difference lies on the launch command, which can be either specified or coded into the container image.

Btw, the elastic mode seems compatible with non-elastic cases.

terrytangyuan: That's good to know.

gaocegege:
Is it still WIP?

zw0610:
> Is it still WIP?

just fixed the test. still trying to add an example as reminded by Terry.

terrytangyuan:
Sounds good. Please remove "WIP" in the PR title and ping me once this is ready for review. Looking forward to it!

pkg/controllers/v1/mpi_job_controller.go
```diff
@@ -684,6 +717,14 @@ func (c *MPIJobController) getOrCreateConfigMap(mpiJob *kubeflow.MPIJob, workerR
 		return nil, fmt.Errorf(msg)
 	}
 
+	// If the ConfigMap is changed, update it
+	if !reflect.DeepEqual(cm.Data, newCM.Data) {
+		cm, err = c.kubeClient.CoreV1().ConfigMaps(mpiJob.Namespace).Update(newCM)
+	}
+	if err != nil {
```
gaocegege, line 724: Can we move the check to the if control block above?

pkg/controllers/v1/mpi_job_controller.go
```diff
@@ -1075,6 +1140,35 @@ shift
 	}
 }
 
+// updateConfigMap
```
gaocegege, line 1143: Please update the comment

zw0610:
> Sounds good. Please remove "WIP" in the PR title and ping me once this is ready for review. Looking forward to it!

@terrytangyuan please take a look.

examples/horovod/tensorflow-mnist-elastic.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  slotsPerWorker: 1
+  cleanPodPolicy: Running
+  mpiReplicaSpecs:
+    Launcher:
+      replicas: 1
+      template:
+        spec:
+          containers:
+          - image: docker.io/kubeflow/mpi-horovod-mnist
```
terrytangyuan, line 14: Is there a Dockerfile for this as a reference? Users might want to customize it.

pkg/controllers/v1/mpi_job_controller.go
```diff
@@ -776,6 +817,30 @@ func (c *MPIJobController) getOrCreateWorker(mpiJob *kubeflow.MPIJob) ([]*corev1
...
+	if err != nil {
+		return nil, err
+	}
+	if len(podFullList) > int(*workerReplicas) {
+		for _, pod := range podFullList {
+			indexStr := strings.TrimLeft(pod.Name, fmt.Sprintf("%s-", workerPrefix))
+			index, err := strconv.Atoi(indexStr)
+			if err == nil {
+				if index >= int(*workerReplicas) {
+					err = c.kubeClient.CoreV1().Pods(pod.Namespace).Delete(pod.Name, &metav1.DeleteOptions{})
```
terrytangyuan, line 835: Do we need to expose delete options?

examples/horovod/tensorflow-mnist-elastic.yaml
terrytangyuan: In particular, they'd be interested in what versions of TF and Horovod they should be using in this image.

terrytangyuan (commented):
Looks great overall! Just left some small comments/questions.

pkg/controllers/v1/mpi_job_controller.go
gaocegege: Do you mean `&metav1.DeleteOptions{}`?

terrytangyuan: Yes

examples/horovod/tensorflow-mnist-elastic.yaml
zw0610: I'm trying to build the docker image with Horovod version jumped to 0.20.0. It failed awkwardly. Given horovod has its own official image, any merits to build one by kubeflow itself?

terrytangyuan: I think it's fine to add a link to their official Dockerfiles/build instructions. We've seen people asking for a Dockerfile before regarding one of the examples in the repo.

pkg/controllers/v1/mpi_job_controller.go
zw0610: I would prefer keep the default delete options before we reached a consensus on how to expose these delete options. If we expose this by adding fields on mpijob crd, that could lead to a bigger modification to the code.

examples/horovod/tensorflow-mnist-elastic.yaml
zw0610: Another way is we simply change the dockerfile.cpu, removing the build phase. In that way, the dockerfile.cpu will only add example scripts to `horovod/horovod:0.20.0-tf2.3.0-torch1.6.0-mxnet1.5.0-py3.7-cpu`.

pkg/controllers/v1/mpi_job_controller.go
gaocegege: I think so.

examples/horovod/tensorflow-mnist-elastic.yaml
gaocegege: I think we can do it in another PR, WDYT

gaocegege:
/lgtm
/approve

I will open the iss

pkg/controllers/v1/mpi_job_controller.go
terrytangyuan: Sounds good. We can add when needed later.

examples/horovod/tensorflow-mnist-elastic.yaml
terrytangyuan: Yes that's fine.

PR #338, opened 2021-03-10
fix --min-np and --max-np args in elastic example

zw0610:
Fix the missing `--` before `min-np` and `max-np`.

### DIFF
--- a/examples/horovod/tensorflow-mnist-elastic.yaml
+++ b/examples/horovod/tensorflow-mnist-elastic.yaml
@@ -18,9 +18,9 @@ spec:
             args:
             - -np
             - "2"
-            - min-np
+            - --min-np
             - "1"
-            - max-np
+            - --max-np
             - "3"
             - --host-discovery-script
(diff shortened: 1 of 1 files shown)


examples/horovod/tensorflow-mnist-elastic.yaml
```diff
@@ -18,9 +18,9 @@ spec:
             args:
             - -np
             - "2"
-            - min-np
+            - --min-np
             - "1"
-            - max-np
+            - --max-np
```
gaocegege, line 23: I am wondering how to support autoscale with HPA in the future. If we change the spec.replicas to 100, what will happen?

terrytangyuan (approved):
Approving this so the example would at least work. We should probably discuss the case of HPA in a separate thread.

/approve
/lgtm

zw0610:
--max-np sets the max processes count. If replicas number exceeds max-np,
redundant slots will be ignored. To support HPA, users needs to give a
large number to max-np in launcher pod command.

PR #340, opened 2021-03-17
MPI Operator: Consolidate manifests

yanniszark:
#### Current manifests structure

```
.
└── mpi-operator
    ├── base
    └── overlays
        └── application
```

#### Explanation

- The whole structure came from kubeflow/manifests.
- The `application` overlays contain Application CRs, which we want to move away from (https://github.com/kubeflow/manifests/issues/1715#issuecomment-779402779)

#### Recommended end state

```
.
└── mpi-operator
    ├── base
    └── overlays
        ├── kubeflow
        └── standalone
```


yanniszark:
cc @kubeflow/wg-training-leads

terrytangyuan (commented):
My main concern is about compatibility of installation methods that we kept changing

yanniszark:
@terrytangyuan this is a valid concern indeed. Let me explain how I have addressed that:
- I have validated the diff of `manifests/mpi-operator/overlays/application` and `manifests/overlays/kubeflow` is the same, except some labels and the application CR.
- I have edited existing docs and scripts to account for the new location.

What do you think?

terrytangyuan (approved):
/lgtm
/approve

PR #334, opened 2021-03-05
fix role updating issue

zw0610:
When worker replicas is scaled up, the rules in the role associated with the launcher pod needs to be update with the new worker pods, enable `horovodrun` to execute command in new worker pods via `kubectl`.

- [x] fix updating rules in role
- [x] fix elastic training example

PR #360, opened 2021-05-17
Propose a new architecture with focus on scalability and robustness

### LINKED ISSUE #315: [Discussion] Switch to use Job API with ArrayJob semantics
From Aldo Culquicondor:
> I left you a message in k8s Slack, but posting it here too in case you are not so active there.
>
>I'm a sig-scheduling maintainer and I'm now looking into supporting array Jobs https://github.com/kubernetes/kubernetes/issues/97169
>We want this to enable MPI workloads on the Job API. I noticed that the MPI operator uses a StatefulSet. That seems like the best idea today. I'm wondering what it would take to migrate to Job, other than having a stable host name.
>
>For extra context, we are also looking into having Job level orchestration for provisioning and scheduling. We have published this doc on how we envision it to work: bit.ly/k8s-job-management
>
>I hope these topics interest you. Feel free to answer here or on Slack,
>

Related issue: https://github.com/kubernetes/kubernetes/issues/97169

cc @alculquicondor @ahg-g @kubeflow/wg-training-leads 

Let's use this issue to discuss with the community.

alculquicondor:
Fixes #315

### DIFF
--- a/proposals/scalable-robust-operator.md
+++ b/proposals/scalable-robust-operator.md
@@ -1,0 +1,284 @@
+# A scalable and robust operator
+
+Authors: @alculquicondor, @ahg-g
+
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Background](#background)
+- [Design](#design)
+- [Alternatives Considered](#alternatives-considered)
+- [Appendix](#appendix-prototype-objects)
+
+## Motivation
+
+A scalable MPI setup on Kubernetes is important for:
+- Running jobs from multiple users in a single cluster.
+- High-performance computing jobs that can require a big number of workers.
+
+A robust MPI setup should be tolerant to failures, implementing retries while
+keeping track of failures.
+
+## Goals
+
+- Allow driver-to-worker control to scale by removing kube-apiserver from
+  the communication channel.
+- Reduce the complexity of the controller by relaying on Kubernetes workload
+  APIs for Pod creation and management.
+  
+## Background
+
+The original design of the operator can be found as a
+[proposal to the kubeflow community](https://github.com/kubeflow/community/blob/master/proposals/mpi-operator-proposal.md).
+The latest release includes a v1alpha2 API and controller.
+A v1 is under development. Something to highlight in this new version is the
+replacement of the Job and StatefulSet by plain Pods, with the intent of
+[tracking running Pods](https://github.com/kubeflow/mpi-operator/issues/201#issuecomment-827837831).
+
+An MPIJob CRD describes the Job. Important fields include:
+- The workers template
+- The number of workers
+- The launcher template, which should have a `mpirun` command.
+
+The images are expected to have the MPI implementation binaries (such as
+OpenMPI, MPICH or Intel MPI) the user’s MPI executable.
+
+A controller processes the MPIJob, starting a Job with the following steps:
+1. Creates ConfigMap, which contains:
+  - A script `kubexec.sh` that wraps `kubectl exec` and is used in replacement
+    of `ssh`. This script, before executing the command provided by `mpirun`,
+    transfers a file containing a mapping of pod names to IPs and appends it to
+    the worker’s `/etc/hosts`.
+    
+    Note: The v1 controller no longer copies the pod-to-IP mapping file. The
+    OpenMPI implementation does the [routing](https://www.open-mpi.org/faq/?category=tcp#tcp-routability-1.3).
+  - The `hostfile` for `mpirun`, listing the worker pod names and number of
+    slots (which could be the number of CPUs or GPUs). This list is built
+    programmatically.
+2. Creates a ServiceAccount+Role+RoleBinding for the launcher, which allow it
+  to:
+  - get/list/watch on Pods
+  - do pods/exec
+  This allows the launcher Pod to obtain details of the worker Pods and start
+  the process managers on them.
+3. If configured, it creates a Volcano PodGroup
+4. Creates a StatefulSet for workers (plain pods in v1). The Pod template
+  includes:
+  - mounts for the ConfigMap.
+  - `sleep` as command.
+5. Creates launcher Job (plain pod in v1). The Pod template includes:
+  - An init container, `kubectl-delivery`, described below.
+  - Environment variables for:
+    - replacing `ssh` for `kubexec.sh`
+    - the `hostfile` location
+  - Volumes for:
+    - The ConfigMap
+    - Sharing files from `kubectl-delivery`
+
+The launcher Job, as previously mentioned, contains an init container:
+`kubectl-delivery`. This is a Kubernetes controller that watches pods. It does
+the following:
+
+1. Copy kubectl from the image into the volume shared with the main container.
+2. Wait for all Pods in the hostfile to be running
+3. Generates a file mapping pod name to IP, in `/etc/hosts` format.
+
+To update the status of an MPIJob, the controller uses the status of the
+launcher Job. That is, when the launcher Job fails or succeeds, it’s status is
+copied to the MPIJob. In v1, it bases the status on the termination condition of
+the Pod.
+
+### Analysis
+
+The above architecture for MPI Jobs has the following problems:
+
+- It doesn’t scale as we increase the number of workers, belonging to the same
+  or different jobs.
+  - Due to the use of `kubectl exec`, every worker spawn goes through
+    kube-apiserver. `mpirun` starts a daemon in each worker
+    (like [`orted`](https://www.open-mpi.org/doc/v3.0/man1/orted.1.php)).
+    This process handles the worker-to-worker communication, which happens
+    without the intervention of kube-apiserver. However, the `exec` connection
+    stays up for control during the entirety of the job.
+  - The `kubectl-delivery` controller does a full cache sync to be able to watch
+    Pods. This startup penalty increases with the number of pods in the cluster
+    and has to be paid for every job. The API calls also causing additional
+    stress on the apiserver as the number of launchers increases.
+  - The launcher pod has execution permissions on any other Pod in the
+    namespace, which can be a security concern.
+  - The v1 controller doesn’t implement launcher pod retries, although there are
+    plans to. So the MPIJob behaves like a plain Pod in this version.
+
+## Design
+
+In order to address the problems of the existing architecture, we propose the
+following changes:
+
+- **The use of `ssh` instead of `kubectl exec`.**
+  
+  This would avoid any pod-to-pod communication happening through apiserver and
+  doesn’t require giving execution permissions at the namespace level.
+  This can be achieved like this:
+  - The controller can generate a single key and share it with the launcher and
+    worker pods through a Secret.
+  - The launcher and workers mount the Secret and set appropriate file
+    permissions.
+  - The workers run an SSH server instead of `sleep`.
+  - When encrypted communication is not a requirement, users could have the
+    choice to use `rsh` for faster communication.
+
+- **The use of stable hostnames and a headless Service for the workers**
+  - This removes the need to query Pod IPs, as the Pods can discover each other
+    through DNS resolution. The hostfile can be generated statically by the
+    controller using the stable hostnames.
+  - Starting with k8s 1.22, we can use
+    [Indexed Jobs with stable hostnames](https://git.k8s.io/enhancements/keps/sig-apps/2214-indexed-job)
+    to delegate the pod management to Kubernetes. Additionally, this will give
+    us robust failure tracking so that we can give users control over retry
+    limits.
+    In the meantime, we can continue using plain Pods.
+    - Caveat 1: The Job controller doesn’t report the number of running Pods.
+      Instead, it reports active Pods, which include running and pending
+      (scheduling or starting). But Kubernetes SIG Apps is
+      [open to add a status field for running Pods](https://kubernetes.slack.com/archives/C18NZM5K9/p1619549430067400).
+    - Caveat 2: Horovod supports elastic workers, but the Kubernetes Job
+      doesn’t support changes to the completions field. This can be supported
+      starting from 1.23. In the meantime, we can replicate the behavior by
+      creating a new Job and doing Pod adoption.
+  - For Intel MPI, we also need a headless Service to front the launcher.
+- **Revert the use of the Job API for the launcher.**
+  - The Job controller handles retries when the launcher or any of the workers fail.
+  - Caveat 1 also applies: The Job controller doesn’t report if the Pod is running.
+    We can continue watching Pods in the meantime.
+- With the above changes, **the following objects can be removed**:
+  - The ServiceAccount+Role+RoleBinding for the launcher.
+  - The `kubectl-delivery` init container in the launcher, as there is no need
+    to obtain IPs, speeding up startup time.
+    
+## Alternatives Considered
+
+TBD from discussions
+
+## Appendix: Prototype objects
+
+It uses a StatefulSet in place of Indexed Jobs, as they are still an alpha
+feature in Kubernetes.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: mpi-config
+data:
+  hostfile: |
+    mpi-workers-0.mpi-workers slots=3
+    mpi-workers-1.mpi-workers slots=3
+    mpi-workers-2.mpi-workers slots=3
+```
+
+```yaml
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/ssh-auth
+data:
+  ssh-privatekey: PRIVATE_KEY
+  ssh-publickey: PUBLIC_KEY
+```
+
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: mpi-launcher
+spec:
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: driver
+        image: '<USER_IMAGE>'
+        args:
+        - 'mpirun'
+        - '-np'
+        - '9'
+        - '<USER_EXECUTABLE>'
+        env:
+        - name: 'OMPI_MCA_orte_keep_fqdn_hostnames'
+          value: 'true'
+        - name: 'OMPI_MCA_orte_default_hostfile'
+          value: '/home/mpiuser/config/hostfile'
+        volumeMounts:
+        - name: ssh-auth
+          mountPath: /mnt/ssh
+          readOnly: true
+        - name: mpi-config
+          mountPath: /home/mpiuser/config
+          readOnly: true
+      volumes:
+      - name: mpi-config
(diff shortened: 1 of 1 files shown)


proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+
+In order to address the problems of the existing architecture, we propose the
+following changes:
+
+- **The use of `ssh` instead of `kubectl exec`.**
+  
+  This would avoid any pod-to-pod communication happening through apiserver and
+  doesn’t require giving execution permissions at the namespace level.
+  This can be achieved like this:
+  - The controller can generate a single key and share it with the launcher and
```
ahg-g, line 121:
```suggestion
  - The controller generates a single key and share it with the launcher and
```

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+    worker pods through a Secret.
+  - The launcher and workers mount the Secret and set appropriate file
+    permissions.
+  - The workers run an SSH server instead of `sleep`.
+  - When encrypted communication is not a requirement, users could have the
```
ahg-g, line 126:
```suggestion
  - When encrypted communication is not a requirement, users have the
```

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+    In the meantime, we can continue using plain Pods.
+    - Caveat 1: The Job controller doesn’t report the number of running Pods.
+      Instead, it reports active Pods, which include running and pending
+      (scheduling or starting). But Kubernetes SIG Apps is
+      [open to add a status field for running Pods](https://kubernetes.slack.com/archives/C18NZM5K9/p1619549430067400).
+    - Caveat 2: Horovod supports elastic workers, but the Kubernetes Job
+      doesn’t support changes to the completions field. This can be supported
+      starting from 1.23. In the meantime, we can replicate the behavior by
+      creating a new Job and doing Pod adoption.
+  - For Intel MPI, we also need a headless Service to front the launcher.
```
ahg-g, line 147: Add: because workers connect back to the launcher.

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+    (like [`orted`](https://www.open-mpi.org/doc/v3.0/man1/orted.1.php)).
+    This process handles the worker-to-worker communication, which happens
+    without the intervention of kube-apiserver. However, the `exec` connection
+    stays up for control during the entirety of the job.
+  - The `kubectl-delivery` controller does a full cache sync to be able to watch
+    Pods. This startup penalty increases with the number of pods in the cluster
+    and has to be paid for every job. The API calls also causing additional
+    stress on the apiserver as the number of launchers increases.
+  - The launcher pod has execution permissions on any other Pod in the
+    namespace, which can be a security concern.
```
ahg-g, line 107: the scale issue here is that to workaround the security concern, each job needs its own namespace, which is not scalable.

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+      creating a new Job and doing Pod adoption.
+  - For Intel MPI, we also need a headless Service to front the launcher.
+- **Revert the use of the Job API for the launcher.**
+  - The Job controller handles retries when the launcher or any of the workers fail.
+  - Caveat 1 also applies: The Job controller doesn’t report if the Pod is running.
+    We can continue watching Pods in the meantime.
+- With the above changes, **the following objects can be removed**:
+  - The ServiceAccount+Role+RoleBinding for the launcher.
+  - The `kubectl-delivery` init container in the launcher, as there is no need
+    to obtain IPs, speeding up startup time.
```
carmark, line 155: Do u mean the init-container `kubectl-delivery` will be removed? If that, how could we  keep the launcher start after the  workers?

proposals/scalable-robust-operator.md
carmark, line 107: The launcher has only the execution permission on the worker pods belong it's job, you can find it at [code](https://github.com/kubeflow/mpi-operator/blob/master/pkg/controllers/v1/mpi_job_controller.go#L1222)

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+To update the status of an MPIJob, the controller uses the status of the
+launcher Job. That is, when the launcher Job fails or succeeds, it’s status is
+copied to the MPIJob. In v1, it bases the status on the termination condition of
+the Pod.
+
+### Analysis
+
+The above architecture for MPI Jobs has the following problems:
+
+- It doesn’t scale as we increase the number of workers, belonging to the same
```
zw0610, line 94: I believe this issue is addressed in v1 controller when we enable horovod elastic work on mpi-operator. https://github.com/kubeflow/mpi-operator/blob/master/proposals/elastic-horovod.md

Besides, I'm not sure what it means 'workers belonging to the same or different jobs'. What is the user case to increase the number of workers for a *different* job?

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+    This process handles the worker-to-worker communication, which happens
+    without the intervention of kube-apiserver. However, the `exec` connection
+    stays up for control during the entirety of the job.
+  - The `kubectl-delivery` controller does a full cache sync to be able to watch
+    Pods. This startup penalty increases with the number of pods in the cluster
+    and has to be paid for every job. The API calls also causing additional
+    stress on the apiserver as the number of launchers increases.
+  - The launcher pod has execution permissions on any other Pod in the
+    namespace, which can be a security concern.
+  - The v1 controller doesn’t implement launcher pod retries, although there are
```
zw0610, line 108: What are user cases for launcher pod retry?

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,284 @@
...
+  or different jobs.
+  - Due to the use of `kubectl exec`, every worker spawn goes through
+    kube-apiserver. `mpirun` starts a daemon in each worker
+    (like [`orted`](https://www.open-mpi.org/doc/v3.0/man1/orted.1.php)).
+    This process handles the worker-to-worker communication, which happens
+    without the intervention of kube-apiserver. However, the `exec` connection
+    stays up for control during the entirety of the job.
+  - The `kubectl-delivery` controller does a full cache sync to be able to watch
+    Pods. This startup penalty increases with the number of pods in the cluster
+    and has to be paid for every job. The API calls also causing additional
```
zw0610, line 104:
```suggestion
    and has to be paid for every job. The API calls also cause additional
```

proposals/scalable-robust-operator.md
ahg-g: Running jobs on a spot/preemptible VMs.

zw0610: In my perspective, I don't think the launcher pod should be scheduled to a spot/preemptible VMs. For the worker pods, yes, because worker pods are stateless. However, the launcher pod seems a stateful instance, whose re-restart means a new job.

proposals/scalable-robust-operator.md
ahg-g: The sentence perhaps needs some rephrasing, I think the point here is that the current setup doesn't scale when increasing the number of workers of a job (e.g., running a 1000 worker pod Job), or increasing the number of jobs (running 100s of MPI jobs). 

The current design puts a lot pressure on central component, the api server:

- `kubectl exec` runs through the API server, and so all jobs running on the same cluster will have to launch their workers using connections established through the api server, which is not scalable. This doesn't scale as we increase the total number of workers across all jobs running on the same cluster.
- pod cache syncs are expensive on large clusters, doing that for each job results in streaming a large amount of state from the api server to the launcher pod. This doesn't scale as we increase the number of jobs.

proposals/scalable-robust-operator.md
alculquicondor: There are 2 options:
1. Do nothing, leave such handling to the gang scheduler. For this reason it is important that the launcher can do retries.
1. To not create the launcher until all the worker pods are running. So everything is handled by the controller, no need for extra cache syncs.

proposals/scalable-robust-operator.md
alculquicondor: Ah, missed that. However, that doesn't scale as the number of workers increase (k8s objects have a size limit).

proposals/scalable-robust-operator.md
alculquicondor: Still, when running the workers in a preemptible VM, if any of them fail, the entire job fails, including the launcher. So we need launcher retries.

alculquicondor (commented):
Thanks for your comments so far. Please let me know if I missed anything else in the background. I got all the details by reading code.

proposals/scalable-robust-operator.md
alculquicondor: According to the other comment, this is not a security issue, as the role has the list of pods. But it's still a scalability issue. See update.

proposals/scalable-robust-operator.md
terrytangyuan: @alculquicondor What MPI implementation are you expecting users to use after launcher gets restarted to make sure the workers are aware and can reconnect to the new launcher?

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,289 @@
...
+  - The `kubectl-delivery` init container in the launcher, as there is no need
+    to obtain IPs, speeding up startup time.
+    
+## Alternatives Considered
+
+TBD from discussions
+
+## Appendix: Prototype objects
+
+It uses a StatefulSet in place of Indexed Jobs, as they are still an alpha
```
terrytangyuan, line 168: Could you add a link to "Indexed Jobs"?

proposals/scalable-robust-operator.md
alculquicondor: The process goes like this:
1. a worker terminates unexpectedly
2. the launcher notices the worker failure, it closes the rest of ssh connections and terminates with a failure
3. the launcher restarts, launches new ssh connections to the workers
This is independent of the MPI implementation.

proposals/scalable-robust-operator.md
carmark: I don't think so, you would like to speed up startup time, but it does not if you want the launcher to retry for failures. And the scheduler should not know about the job startup strategy.

carmark: By the same time, how do you verify the  real launcher failed reason for option 1? for example,
* workers does not startup, the launcher failed and restart
* training job failed, the launcher failed.

proposals/scalable-robust-operator.md
zw0610: > Ah, missed that. However, that doesn't scale as the number of workers increase (k8s objects have a size limit).

It does. We introduced this feature in supporting Horovod Elastic: https://github.com/kubeflow/mpi-operator/pull/334

proposals/scalable-robust-operator.md
zw0610: > Still, when running the workers in a preemptible VM, if any of them fail, the entire job fails, including the launcher. So we need launcher retries.

From what I observed from other HPC system, like Slurm, for jobs that are not fault-tolerant, when some of the worker fails, it does not retry via a launcher restart. Instead, the entire job will be marked as failed with it resources released. The system will re-queue the failed job if 'retry' is demanded by the user and create a new job. Such process is compatible with the contemporary design of mpi-operator.

proposals/scalable-robust-operator.md
gaocegege: If we replace kubectl exec with ssh, we can also support the elastic feature, thus I do not think it is a problem for us if I understand it correctly.

proposals/scalable-robust-operator.md
```diff
@@ -0,0 +1,289 @@
...
+  name: mpi-workers
+spec:
+  clusterIP: None
+  selector:
+    app: mpi-workers
+```
+
+```yaml
+apiVersion: apps/v1
+kind: StatefulSet
```
gaocegege, line 250: I am wondering how to support elastic mode with statefulset? We may add/remove workers in the fly.

proposals/scalable-robust-operator.md
alculquicondor: What do you mean by "job startup strategy"?

As for differentiating failures, I don't think we can or should do that. There is another type of failure which is kind of a mix of the two you described: a worker goes down. In this case, the job might have already started.

But perhaps option 2 is reasonable. WDYT?

proposals/scalable-robust-operator.md
alculquicondor: You just change `.spec.replicas`.

proposals/scalable-robust-operator.md
alculquicondor: OTOH, it kind of goes against the declarative and fault-tolerant approach of k8s APIs, including the Job API. Retries is what kubernetes users would expect. And if they don't need it, they could always disable it.

alculquicondor:
There are 2 main open discussions pending:
- Should we hold the creation of the launcher until the workers are all ready? https://github.com/kubeflow/mpi-operator/pull/360#discussion_r634211580
- Are retries necessary? https://github.com/kubeflow/mpi-operator/pull/360#discussion_r634222943

Please leave your thoughts

proposals/scalable-robust-operator.md
ahg-g: I think it is useful if we make automatic retries an option that we default to 0.

proposals/scalable-robust-operator.md
ahg-g: If we want to support retries, then by product we want option 1, correct?

alculquicondor: we can still use option 2. It might make startups less flaky. But it's not a guarantee anyways. A worker can simply just fail in between the time it reported running and the launcher started running.

alculquicondor:
I asked for 10 min in the next AutoML and Trainging WG meeting. Would you be able to make it @terrytangyuan @gaocegege ?

terrytangyuan:
I cannot make it due to scheduling conflicts. In the meantime, I'd encourage all existing reviewers to check if your comments can be resolved or discuss if there are further concerns on the proposal (which might be more efficient as we are located in more than 4 timezones).

ahg-g:
> I cannot make it due to scheduling conflicts. In the meantime, I'd encourage all existing reviewers to check if your comments can be resolved or discuss if there are further concerns on the proposal (which might be more efficient as we are located in more than 4 timezones).

+1. Can we also assign an "approver"?

terrytangyuan:
We should try to reach consensus from people listed in [OWNERS](https://github.com/kubeflow/mpi-operator/blob/master/OWNERS). People from @kubeflow/wg-training-leads can also approve after that.

alculquicondor:
@rongou any thoughts?

gaocegege:
@carmark @zw0610 @Jeffwan PTAL thanks.

alculquicondor:
One topic that was raised at today's meeting was where to place the code. This question is important because there are breaking changes and the new controller wouldn't be able to properly process an existing job.

I was thinking that we can just modify v1 code, as it hasn't been released, AFAIK. Is this not the case? Should we create a new folder? like `v1-ssh` or something like that? My worry is that this would become a maintenance problem, as now changes would have to be applied in 2 places. And long term, the existing v1 code should be removed. But if v1 was never released, it should be safe to just replace it.

terrytangyuan:
It's true that v1 is not officially released yet but I believe there are several companies that are already running it in production which might be concerning.

alculquicondor:
I see. Then maybe we can create the new controller in a entire new module, so that we can start with fresh dependencies and not be blocked by kube-batch's (see #364)

> We should try to reach consensus from people listed in OWNERS. People from @kubeflow/wg-training-leads can also approve after that.

If they don't respond, can we assume lazy consensus?

zw0610:
> It's true that v1 is not officially released yet but I believe there are several companies that are already running it in production which might be concerning.

I believe the API version `v1`, `v1alpha1` and `v1alpha2` is not necessary bounded to the controller version, which means, if this proposal wish to add a kind of new controller to the `v1` API, all the contributors need to do is add the implementation under `pkg/controllers/v1` but in a different file like `pkg/controllers/v1/mpi_job_alternative_controller.go`. After the implementation is done, you can add option in `cmd/mpi-operator.v1` to let user choose which controller to use. In this way, we can avoid users who are already using the `v1` API can the existing controller implementation.

alculquicondor:
The problem is not necessarily the API. Let's say someone has running jobs with the existing v1 controller. If they upgrade to the new proposed controller, the existing jobs would have orphan resources that the new controller won't manage. And the existing workers wouldn't work, because they don't have stable hostnames.

alculquicondor:
In any case, I think the decision of where to place the code comes later. Can we have an agreement on the direction of the new architecture? See https://github.com/kubeflow/mpi-operator/pull/360#issuecomment-845445542

alculquicondor:
Thanks, I'll work on that throughout the week.

alculquicondor:
I ran some experiments with these characteristics:

- 21 nodes
- 600 running pods (not part of the job)
- Job with 3 slots per worker
- All nodes already have the image (so no image pull time involved)

Then I run two jobs, one with 3 workers and one with 20 workers. Each task is doing:
1. print its rank
2. Execute a simple calculation
3. Do `MPI_Reduce` targeting rank 0

So really, most of the time is spent on setup, but some communication is being tested too.

For each run, I recorded the Start time and Completion time as reported by the mpi-operator controller.

For 3 workers, I got an average of 9s for the current operator and 3s for the ssh proposal.
For 20 workers, I got an average of 19s for the current operator and 11s for the ssh proposal.

All runs here:

image

You can see an important improvement. I used a reasonably small cluster and somewhat small jobs. Let me know if you would like any more details about my quick experiments.

However, let me reiterate on theoretical analysis already presented, which I think is important on its own. The apiserver is a critical component in the cluster. We should avoid stressing it with requests that intend to do pod-to-pod communication. The apiserver has limits for ongoing open connections and it throttles API requests. With the current architecture, just a few dozens of jobs could cause starvation to API requests from critical components such as the scheduler or the controller manager.

alculquicondor:
@terrytangyuan @zw0610 anything to add?

terrytangyuan:
Thanks for putting this together! I think both theoretical and analytical improvements look great to me. 

As it's already close to the weekend, **please leave additional comments by June 16th and I will merge this if there's no major concerns by then.**

johnugeorge:
I feel that this proposal is a strong case for a revamp based on https://github.com/kubeflow/mpi-operator/pull/360#issuecomment-861632258. 

/lgtm

gaocegege:
/lgtm

Thanks for the proposal.

terrytangyuan (approved):

Issue #359, opened 2021-05-11
failed to go get github.com/kubeflow/mpi-operator (invalid version: unknown revision 000000000000)


heyfey:
Hi, 
I want to use clienset in my project, therefore, I need to import it.

When I tried `go get github.com/kubeflow/mpi-operator`
I got
```
go: github.com/kubeflow/mpi-operator@v0.2.3 requires
        k8s.io/sample-controller@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

Then, I try to include it into my `go.mod` :
```
module github.com/XXX/XXX

go 1.13

require github.com/kubeflow/mpi-operator v0.2.3
```
and use `go mod tidy`. The result is the same.

I've tried go 1.16 and 1.13, both got same results.
Any suggestions?

carmark:
sure, I will fix it.

PR #361, opened 2021-05-19
update gomod for sample-controller

carmark:
fix https://github.com/kubeflow/mpi-operator/issues/359


go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	k8s.io/client-go v10.0.0+incompatible
 	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20200410163147-594e756bea31
-	k8s.io/sample-controller v0.0.0-00010101000000-000000000000
+	k8s.io/sample-controller v0.16.15
```
gaocegege, line 19: Why do we need sampler controller?

terrytangyuan: Is there a way to get rid of it? Feels a little weird to rely on it.

terrytangyuan (approved):
Merging for now to unblock the existing issue. Let's track the issue with sample controller separately.

/approve

Issue #364, opened 2021-06-02
Upgrade mod dependencies


alculquicondor:
Currently, the operator builds against the 1.15 k8s libraries. Note that currently the oldest supported version of k8s is 1.19

Another dependency is kubeflow/common 0.3.3, which builds against k8s 1.16 libraries. 0.3.4 builds against 1.19

So we could at least upgrade to 1.19 dependencies. This should make it simpler to introduce more dependencies such as controller-runtime for integration and E2E testing #233

However, there is one problematic dependency: kube-batch. While volcano has builds against k8s 1.19, kube-batch (deprecated project) only has builds against 1.16. This dependency is used in the v1alpha2 controller. Hopefully we can still build it using "replace" directives. It might be possible given that we only use the listers from it.

alculquicondor:
cc @terrytangyuan

terrytangyuan:
Possibly related: https://github.com/kubeflow/mpi-operator/issues/323

alculquicondor:
So I gave it a try, and it seems that volcano is fine (it has a version that builds against 1.19).

The problem is kube-batch, used in v1alpha2. I suppose it's not safe to migrate v1alpha2 from kube-batch to volcano, right?

alculquicondor:
I tried to upgrade to 1.18, 1.17 and even 1.16. In all of them kube-batch fails to build.

Since this dependency nightmare might limit our abilities to iterate the v1 controller, I propose we move it to a separate module https://blog.golang.org/v2-go-modules

Or maybe I can just have the new controller proposed in #360 in the new module, solving the question about keeping the old and new one.

alculquicondor:
The dependencies were upgraded for the v2 controller #370

/close

PR #366, opened 2021-06-21
Fork v2 controller and API in a new module

alculquicondor:
A new module will host the implementation of #360

This PR copies all files from the v1 controller and API without any modifications beyond import paths.

The new module will allow to handle the requirements independently from older controllers https://github.com/kubeflow/mpi-operator/issues/364#issuecomment-853921141

Also forked the Makefile (inside the new module folder) and the Dockerfile (outside). Changed update-codegen.sh to handle the new module.

A follow up PR will update the k8s dependencies and then the architectural modifications will come.


terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #368, opened 2021-06-22
Fix kubectl-delivery security issues

ryantd:
Fix gosec G304 G307 security issues

Ref
[G304: File path provided as taint input](https://securego.io/docs/rules/g304.html)
[G307: Unsafe defer call of a method returning an error for defer file.Close() ?](https://github.com/securego/gosec/issues/512#issuecomment-675286833)

PR #369, opened 2021-06-22
Merge v2 Makefile into main one

alculquicondor:
This removes the need for a separate Dockerfile


terrytangyuan (commented):
Is there anything else that can be merged into one as well to reduce maintenance efforts?

terrytangyuan:
What's concerning is that changes won't be reflected in the fork, e.g. https://github.com/kubeflow/mpi-operator/pull/368

alculquicondor:
I don't think we can merge anything else. #368 doesn't actually matter for the new controller (kubectl-delivery is going away)

terrytangyuan (approved):
Thanks!

/lgtm
/approve

Issue #373, opened 2021-07-06
Implement v2 controller that sets up SSH for communication


alculquicondor:
Implementation for https://github.com/kubeflow/mpi-operator/blob/master/proposals/scalable-robust-operator.md

- [ ] Replace kubectl and support roles with ssh and secrets
- [ ] Ensure the launcher starts after the workers are ready
- [ ] Ensure all container image examples support the SSH setup
- [ ] Replace the plain pod workers with StatefulSet or Indexed Job
- [ ] Replace the plain pod launcher with a Job

gaocegege:
https://www.kubeflow.org/docs/about/contributing/#joining-the-kubeflow-github-org

Hi, could you please join the kubeflow org? Then we do not need to trigger the CICD for your PR manually.

alculquicondor:
Sent PR kubeflow/internal-acls#473

Thanks for the suggestion

alculquicondor:
I verified that images `docker.io/kubeflow/mpi-horovod-mnist` and `docker.io/mpioperator/tensorflow-benchmarks` just work with the new controller. Marking that as done.

Jeffwan:
@alculquicondor Has community discussed tradeoffs about job vs pod for launcher, statefulsets vs plain pods for workers?

alculquicondor:
Yes for launcher. See the discussion here #386 

For workers, it's still open for discussion. We could do Statefulsets, but I think plain pods might be fine for now. We might migrate to Indexed Jobs at some point, but since it's only available in k8s 1.22, it's kind of early to discuss.

alculquicondor:
I think this is pretty much ready. The last things I would like to do are:
- update the labels to use fully-qualified names. For that, I'm waiting for a release of kubeflow/common#153
- Add documentation (is there a website, or should I just do it on readmes)?

terrytangyuan:
>     * Add documentation (is there a website, or should I just do it on readmes)?
There's this page https://www.kubeflow.org/docs/components/training/mpi/

tenzen-y:
Maybe we can introduce Indexed Job to mpi-operator v2 once https://github.com/kubernetes/enhancements/issues/3715 is graduated to beta.

tenzen-y:
Consider introducing JobSet instead of managing raw pods for the workers: https://github.com/kubernetes-sigs/jobset

PR #374, opened 2021-07-06
Revert horovod example to v1

alculquicondor:
The controller generates keys and mounts them to the containers. The container images must know how to place the credentials and set file permissions.

First part of #373 (Replace kubectl and support roles with ssh and secrets)

Tested with the horovod CPU image

### DIFF
--- a/examples/horovod/tensorflow-mnist.yaml
+++ b/examples/horovod/tensorflow-mnist.yaml
@@ -1,4 +1,4 @@
-apiVersion: kubeflow.org/v2
+apiVersion: kubeflow.org/v1
 kind: MPIJob
 metadata:
   name: tensorflow-mnist
@@ -15,6 +15,7 @@ spec:
             name: mpi-launcher
             command:
             - mpirun
+            args:
             - -np
             - "2"
             - --allow-run-as-root


manifests/overlays/dev/kustomization.yaml
```diff
@@ -11,7 +11,7 @@ commonLabels:
   app.kubernetes.io/component: mpijob
 images:
 - name: mpioperator/mpi-operator
-  newName: gcr.io/acondor-gke-dev/mpi-operator
+  newName: dev-registry/mpi-operator
```
terrytangyuan, line 14: Is this change intentional? Does this exist?

alculquicondor: Neither exist (my GCR is private). It's intended for developers to use their preferred registry.

terrytangyuan (approved):
LGTM. Others PTAL

examples/horovod/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
gaocegege, line 1: I am wondering if it is needed for every MPIJob. Should we add some comments about why to introduce such a entrypoint script?

alculquicondor: Yes, something like this will always be needed.
The problem is that SSH is very picky about not just file permissions, but also the permissions of the directory. Unfortunately, kubernetes doesn't allow to change the permissions of the mount folder.

I'll update the README once I'm done with all development (tests included). The idea is: the controller will mount the keys in /mnt/ssh. It's up to the container image to set those files in the appropriate location for their user.

alculquicondor: FTR, I tried to use mount the files in /root/.ssh or used them from /mnt/ssh by setting `AuthorizedKeysFile` and `IdentityFile` in the ssh configuration. However, SSH complained about the folder permissions :(

alculquicondor: Hold on, I'm investigating if I can do the same in an init container.

alculquicondor: I'm surprised, but it worked. Now, I'm not convinced that it will work if the containers don't run as root, but I'll check that later. I added a bullet point in #373.

alculquicondor:
Btw, I found a bug where the incorrect scheme was being used. As a consequence, defaults were not applied and events fail to record. See latest commit.

alculquicondor:
@carmark @Jeffwan any other comments?

examples/horovod/tensorflow-mnist.yaml
```diff
@@ -15,7 +15,6 @@ spec:
             name: mpi-launcher
             command:
             - mpirun
-            args:
```
gaocegege, line 18: Do we need to remove it?

alculquicondor: No, I just messed it up when reverting. Also, using a the v1 api works just fine. the apiserver does the conversion for us.

gaocegege (commented):
PTAL

This PR removes the CLI argument `kubectl-delivery-image`, does it work for you? @carmark @terrytangyuan

gaocegege (commented):
LGTM

terrytangyuan (commented):
LGTM. @zw0610 take another look?

alculquicondor:
uhm.... #375 was already merged, which contain these changes... should we revert? @terrytangyuan

terrytangyuan:
Then we can close this one?

alculquicondor:
I wasn't sure if there were any changes here that I didn't put in the other branch. But it looks like it was all there.

alculquicondor:
This commit was missing

terrytangyuan:
Looks like there’s only one changed file?

alculquicondor:
correct

terrytangyuan:
Could you update the PR title then?

alculquicondor:
Is there a way to retrigger the test? It looks like the event watcher is not working or something. I can't reproduce locally.

terrytangyuan:
Just re-triggered it

PR #378, opened 2021-07-16
Downgrade v2 API to v2beta1

alculquicondor:
Separate from the maturity of the controller, we should consider the maturity of the API.

I believe our current API has room for improvement. So I'd like to state this in the version name. This should be fine as the new API hasn't been announced as released and I would assume no one has been using it in production yet.


terrytangyuan (approved):
Thanks!

/lgtm
/approve

alculquicondor:
I forgot to change one import alias

alculquicondor:
not sure why coverage decreased with this. Can this be bypassed?

PR #379, opened 2021-07-19
Upgrade to apiextensions.k8s.io/v1

alculquicondor:
- Define all fields but the PodTemplateSpec.
- Remove v1alpha1.
- Remove unused API fields from v2beta1 (`mainContainer`, `runPolicy`)

Ensure Pod created condition is applied to the Job copy.

### DIFF
--- a/v2/pkg/apis/kubeflow/v2beta1/types.go
+++ b/v2/pkg/apis/kubeflow/v2beta1/types.go
@@ -51,15 +51,6 @@ type MPIJobSpec struct {
 	// `MPIReplicaSpecs` contains maps from `MPIReplicaType` to `ReplicaSpec` that
 	// specify the MPI replicas to run.
 	MPIReplicaSpecs map[MPIReplicaType]*common.ReplicaSpec `json:"mpiReplicaSpecs"`
-
-	// MainContainer specifies name of the main container which
-	// executes the MPI code.
-	MainContainer string `json:"mainContainer,omitempty"`
-
-	// `RunPolicy` encapsulates various runtime policies of the distributed training
-	// job, for example how to clean up resources and how long the job can stay
-	// active.
-	RunPolicy *common.RunPolicy `json:"runPolicy,omitempty"`
 }
 
 // MPIReplicaType is the type for MPIReplica.
--- a/v2/pkg/controller/mpi_job_controller.go
+++ b/v2/pkg/controller/mpi_job_controller.go
@@ -459,6 +459,13 @@ func (c *MPIJobController) syncHandler(key string) error {
 		}
 	}
 
+	if len(mpiJob.Status.Conditions) == 0 {
+		msg := fmt.Sprintf("MPIJob %s/%s is created.", mpiJob.Namespace, mpiJob.Name)
+		updateMPIJobConditions(mpiJob, common.JobCreated, mpiJobCreatedReason, msg)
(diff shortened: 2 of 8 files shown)


v2/pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -51,15 +51,6 @@ type MPIJobSpec struct {
...
 	MPIReplicaSpecs map[MPIReplicaType]*common.ReplicaSpec `json:"mpiReplicaSpecs"`
-
-	// MainContainer specifies name of the main container which
-	// executes the MPI code.
-	MainContainer string `json:"mainContainer,omitempty"`
-
-	// `RunPolicy` encapsulates various runtime policies of the distributed training
-	// job, for example how to clean up resources and how long the job can stay
-	// active.
-	RunPolicy *common.RunPolicy `json:"runPolicy,omitempty"`
```
gaocegege, line 62: I am wondering if we should delete these fields.

alculquicondor: They are not used, why have them?

kawych: Is there a design for the CRD (i.e. with explanation how these fields are meant to be used)? Do we understand that the initial reasons for adding those no longer apply?

kawych (commented):
Looks good in general, just a few questions to help me get a better understanding.

v2/pkg/apis/kubeflow/v2beta1/types.go
alculquicondor: `MainContainer` was added to be able to start binaries using `kubectl exec`, as this requires a target container when there are more than one. This is no longer needed, as the communication happens through the SSH port.

`RunPolicy` was used in the v1alpha2 controller to configure the launcher Job retries. This is no longer used in the newer versions as we are managing plain Pods. I think we can bring it back once we move back to using Jobs. But that might take a while.

v2/pkg/controller/mpi_job_controller.go
```diff
@@ -459,6 +459,13 @@ func (c *MPIJobController) syncHandler(key string) error {
 		}
 	}
 
+	if len(mpiJob.Status.Conditions) == 0 {
```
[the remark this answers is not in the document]
alculquicondor: Problems with the previous approach:

- We were modifying an object in the informer cache, instead of modifying a copy. That can lead to undefined behavior.
- If the first Job status update failed, we might never add the condition again.
- If the job controller restarts, we were resetting the condition timestamp and incrementing the mpiJobscreateCount metric for all existing Jobs. When the controller restarts, every job produces an Add event.

v2/pkg/apis/kubeflow/v2beta1/types.go
alculquicondor: To clarify, if we release a v1beta2 API with a field `RunPolicy` that has no effect, and then we release a new version were it suddenly has effects, that's a breaking change.
It's better to release only with the fields that have effects and add the new fields when the implementation for them is also ready.

kawych (approved):

alculquicondor:
@terrytangyuan anything else for approval?

terrytangyuan (approved):
Thanks!

/lgtm
/approve

alculquicondor:
It looks like the coverage was still active when I submitted this. Could you retrigger the checks?

alculquicondor:
@terrytangyuan could you merge manually?

PR #380, opened 2021-07-19
build: Remove code coverage report

terrytangyuan:
This has not been useful for us and we have to manually merge PRs sometimes due to inaccurate report.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,3 @@ jobs:
         run: make mpi-operator.v1alpha1 mpi-operator.v1alpha2 mpi-operator.v1 mpi-operator.v2 kubectl-delivery
       - name: Run tests
         run: make test
-      - name: Send the coverage output
-        uses: shogo82148/actions-goveralls@v1
-        with:
-          path-to-profile: profile.cov


alculquicondor:
I'm actively working on increasing the coverage. I would appreciate if we can keep this. Perhaps we can tune how much lower a change can go?

terrytangyuan:
Perhaps we can just print the coverage results? This has been flaky even for simple changes like a one-line documentation update.

.github/workflows/main.yml
```diff
@@ -21,7 +21,3 @@ jobs:
         run: make mpi-operator.v1alpha1 mpi-operator.v1alpha2 mpi-operator.v1 mpi-operator.v2 kubectl-delivery
       - name: Run tests
         run: make test
-      - name: Send the coverage output
-        uses: shogo82148/actions-goveralls@v1
-        with:
-          path-to-profile: profile.cov
```
alculquicondor, line 27: It looks like the v2 coverage is in a separate file: `v2/profile.cov`

Can we send both? Otherwise, since you are fine removing the coverage report, we can just have the one for v2.

terrytangyuan: I am not sure how to do that in this github actions step. There seems to be a lack of documentation on this. Perhaps when you are increasing the coverage, you can also add a command to print out the coverage locally instead of using coveralls, etc. so this PR would focus on removing the coveralls service.

alculquicondor: The point of the report is not necessarily for my own development. I know I'm increasing coverage. The point is to prevent others from significantly reducing coverage.
But if this particular tool is becoming a burden more than help, then I'm fine with removing it.

terrytangyuan: We can revisit and use a more reliable tool later.

PR #381, opened 2021-07-21
fix comment typo related to statefulset

HeGaoYuan:
Hi, mpi-operator team
    I find that the v1 mpijob controller does not create statefulset anymore, but some comment is very confusing. So I fix it.


HeGaoYuan:
Close it and reopen for some reason. :(

PR #383, opened 2021-07-21
Allow running MPI applications as non-root

alculquicondor:
Part of #373 

- Adds the spec field sshAuthMountPath for MPIJob.
- The init script sets the permissions and ownership based on the securityContext of the launcherPod
- Adds a pure-MPI example to show how to create and use an image as non-root

Tested: unit and integration suites as well as manual runs.

### DIFF
--- a/examples/pi/Dockerfile
+++ b/examples/pi/Dockerfile
@@ -1,0 +1,20 @@
+FROM debian:buster
+
+RUN apt update && apt install -y --no-install-recommends \
+			build-essential \
+			libopenmpi-dev \
+			openmpi-bin \
+			openssh-server \
+			openssh-client \
+			&& rm -rf /var/lib/apt/lists/*
+# Add priviledge separation directoy to run sshd as root.
+RUN mkdir -p /var/run/sshd
+# Add capability to run sshd as non-root.
+RUN setcap CAP_NET_BIND_SERVICE=+eip /usr/sbin/sshd
+
+RUN useradd -m mpiuser
+WORKDIR /home/mpiuser
+COPY --chown=mpiuser sshd_config .sshd_config
+RUN sed -i 's/[ #]\(.*StrictHostKeyChecking \).*/ \1no/g' /etc/ssh/ssh_config
+COPY pi.cc /src/pi.cc
+RUN mpic++ /src/pi.cc -o /home/mpiuser/pi
--- a/v2/pkg/apis/kubeflow/v2beta1/default.go
+++ b/v2/pkg/apis/kubeflow/v2beta1/default.go
@@ -56,6 +56,9 @@ func SetDefaults_MPIJob(mpiJob *MPIJob) {
 	if mpiJob.Spec.SlotsPerWorker == nil {
 		mpiJob.Spec.SlotsPerWorker = newInt32(1)
 	}
+	if mpiJob.Spec.SSHAuthMountPath == "" {
+		mpiJob.Spec.SSHAuthMountPath = "/root/.ssh"
+	}
 
 	// set default to Launcher
 	setDefaultsTypeLauncher(mpiJob.Spec.MPIReplicaSpecs[MPIReplicaTypeLauncher])
--- a/v2/pkg/apis/kubeflow/validation/validation.go
+++ b/v2/pkg/apis/kubeflow/validation/validation.go
@@ -46,6 +46,9 @@ func validateMPIJobSpec(spec *kubeflow.MPIJobSpec, path *field.Path) field.Error
 	} else if !validCleanPolicies.Has(string(*spec.CleanPodPolicy)) {
 		errs = append(errs, field.NotSupported(path.Child("cleanPodPolicy"), *spec.CleanPodPolicy, validCleanPolicies.List()))
 	}
+	if spec.SSHAuthMountPath == "" {
+		errs = append(errs, field.Required(path.Child("sshAuthMountPath"), "must have a mount path for SSH credentials"))
+	}
 	return errs
 }
 
--- a/v2/pkg/controller/mpi_job_controller.go
+++ b/v2/pkg/controller/mpi_job_controller.go
@@ -73,22 +73,21 @@ const (
 	sshAuthSecretSuffix     = "-ssh"
 	sshAuthVolume           = "ssh-auth"
 	sshAuthMountPath        = "/mnt/ssh"
+	sshHomeInitMountPath    = "/mnt/home-ssh"
 	sshHomeVolume           = "ssh-home"
-	// TODO(alculquicondor): Make home directory configurable through the API.
-	sshHomeMountPath       = "/root/.ssh"
-	launcher               = "launcher"
-	worker                 = "worker"
-	launcherSuffix         = "-launcher"
-	workerSuffix           = "-worker"
-	gpuResourceNameSuffix  = ".com/gpu"
-	gpuResourceNamePattern = "gpu"
-	labelGroupName         = "group-name"
-	labelMPIJobName        = "mpi-job-name"
-	labelMPIRoleType       = "mpi-job-role"
-	sshPublicKey           = "ssh-publickey"
-	sshPrivateKeyFile      = "id_rsa"
-	sshPublicKeyFile       = sshPrivateKeyFile + ".pub"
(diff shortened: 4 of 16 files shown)


alculquicondor:
/hold
Does this work?

terrytangyuan:
Yes it should work

v2/pkg/controller/mpi_job_controller.go
```diff
@@ -74,21 +74,19 @@ const (
 	sshAuthVolume           = "ssh-auth"
 	sshAuthMountPath        = "/mnt/ssh"
 	sshHomeVolume           = "ssh-home"
-	// TODO(alculquicondor): Make home directory configurable through the API.
```
kawych, line 77: Just a thing to consider: would it make sense to make the whole home directory configurable as opposed to just .ssh directory?

v2/pkg/controller/mpi_job_controller_test.go
```diff
@@ -889,7 +890,386 @@ func TestWorkerReady(t *testing.T) {
...
+						"mpi-job-role": "launcher",
+					},
+				},
+				Spec: corev1.PodSpec{
+					Hostname:      "foo-launcher",
+					Subdomain:     "foo-worker",
+					RestartPolicy: corev1.RestartPolicyNever,
+					Containers: []corev1.Container{
+						{
+							Env: func() []corev1.EnvVar {
```
kawych, line 941: Can all of this be replaced by "append(ompiEnvVars, nvidiaDisableEnvVars...)?

kawych (commented):
Looks good so far, but it's a fairly large change and it's easy to miss things. Could you confirm that it was tested end-to-end in addition to unit tests?

examples/pi/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+			&& rm -rf /var/lib/apt/lists/*
+# Add priviledge separation directoy to run sshd as root.
+RUN mkdir -p /var/run/sshd
+# Add capability to run sshd as non-root.
+RUN setcap CAP_NET_BIND_SERVICE=+eip /usr/sbin/sshd
+
+RUN useradd -m mpiuser
+WORKDIR /home/mpiuser
+COPY --chown=mpiuser sshd_config .sshd_config
+RUN cat /etc/ssh/ssh_config | grep -Ev 'StrictHostKeyChecking' > /etc/ssh/ssh_config.new && \
```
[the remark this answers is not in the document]
alculquicondor: Simplified it a bit differently, but thanks for the suggestion :)

v2/pkg/controller/mpi_job_controller.go
alculquicondor: I thought about that. But someone creating a container image might want not to use `$HOME/.ssh` but a completely different location. And at the moment I can think of other uses of the home folder from the controller perspective.

v2/pkg/controller/mpi_job_controller_test.go
alculquicondor: I considered it, but not sure if it will have consequences on the allocation of `ompiEnvVars`

another alternative would be:

```
append(append([]corev1.EnvVar(nil), ompiEnvVars...), nvidiaDisableEnvVars...)
```

WDYT?

alculquicondor (commented):
I added a note in the description on what I did to test the change.

v2/pkg/controller/mpi_job_controller_test.go
kawych: Yeah, I've just played with append a little and it's a little funny. [1]: "If it has sufficient capacity, the destination is resliced to accommodate the new elements. If it does not, a new underlying array will be allocated."

[1] https://pkg.go.dev/builtin#append

So in this case "append(ompiEnvVars, nvidiaDisableEnvVars...)" will work the same as "append(append(...)", but if you allocate additional memory for the ompiEnvVars slice, it will have a different effect. Examples:

`a := []int{1}
b := append(a[:1], []int{3}...)
a[0] = 2
b[0] = 3
fmt.Printf("%d %d\n", a[0], b[0])` <- this prints "2 3"

`a := []int{1, 2}
b := append(a[:1], []int{3}...)
a[0] = 2
b[0] = 3
fmt.Printf("%d %d\n", a[0], b[0])` <- this prints "3 3"

So append(append([]corev1.EnvVar(nil), ompiEnvVars...), nvidiaDisableEnvVars...) would be better here.

kawych (approved):

v2/pkg/controller/mpi_job_controller_test.go
alculquicondor: Yeah, that will be more predictable. I decided to add a helper method.

v2/pkg/apis/kubeflow/v2beta1/default.go
```diff
@@ -56,6 +56,9 @@ func SetDefaults_MPIJob(mpiJob *MPIJob) {
 	if mpiJob.Spec.SlotsPerWorker == nil {
 		mpiJob.Spec.SlotsPerWorker = newInt32(1)
 	}
+	if mpiJob.Spec.SSHAuthMountPath == "" {
+		mpiJob.Spec.SSHAuthMountPath = "/root/.ssh"
```
zw0610, line 60: could you make the default path (`"/root/.ssh"`) a constant?

v2/pkg/apis/kubeflow/validation/validation.go
```diff
@@ -46,6 +46,9 @@ func validateMPIJobSpec(spec *kubeflow.MPIJobSpec, path *field.Path) field.Error
 	} else if !validCleanPolicies.Has(string(*spec.CleanPodPolicy)) {
 		errs = append(errs, field.NotSupported(path.Child("cleanPodPolicy"), *spec.CleanPodPolicy, validCleanPolicies.List()))
 	}
+	if spec.SSHAuthMountPath == "" {
```
zw0610, line 49: does this validation contradict with the [defaulting here](https://github.com/kubeflow/mpi-operator/pull/383/files#diff-e4f5a918268e7d4cc61f43c32e525cfb9ff706c0a57ecc0372a2259b22d44c09R59)?

v2/pkg/controller/mpi_job_controller.go
```diff
@@ -1412,61 +1413,58 @@ func workerReplicas(job *kubeflow.MPIJob) int32 {
...
+	initScript := "" +
+		"cp -RL /mnt/ssh/* /mnt/home-ssh && " +
+		"chmod 700 /mnt/home-ssh && " +
+		"chmod 600 /mnt/home-ssh/*"
+	if launcherSecurityCtx != nil && launcherSecurityCtx.RunAsUser != nil {
+		initScript += fmt.Sprintf(" && chown %d -R /mnt/home-ssh", *launcherSecurityCtx.RunAsUser)
+	}
+	podSpec.InitContainers = append(podSpec.InitContainers, corev1.Container{
+		Name:  "init-ssh",
+		Image: "alpine:3.14",
```
zw0610, line 1454: could we make this image configurable? some cluster environment is isolated from docker registry.

v2/pkg/controller/mpi_job_controller.go
```diff
@@ -1412,61 +1413,58 @@ func workerReplicas(job *kubeflow.MPIJob) int32 {
...
+		VolumeMounts: []corev1.VolumeMount{
 			{
 				Name:      sshAuthVolume,
 				MountPath: sshAuthMountPath,
-			}, {
+			},
+			{
 				Name:      sshHomeVolume,
-				MountPath: sshHomeMountPath,
+				MountPath: "/mnt/home-ssh",
```
zw0610, line 1462: why do we replace constant with hardcoded string here?

zw0610 (commented):
I believe this pr does more than allowing non-root mode. If so, I would prefer this pr comes with multiple commits. What do you think? @terrytangyuan

v2/pkg/controller/mpi_job_controller.go
alculquicondor: Done

alculquicondor (commented):
I put the sample in a separate commit

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #385, opened 2021-07-26
Validate that MPIJob produces valid hostnames

alculquicondor:
Hostnames must be valid DNS labels. This includes checking for invalid characters and a maximum length.

Here we check that the name of the MPIJob would produce a valid worker hostname for the biggest possible index.


terrytangyuan (approved):
Thanks!

/lgtm
/assign @gaocegege @zw0610

PR #389, opened 2021-07-28
Add support for Intel MPI

alculquicondor:
- Adds the field `.spec.mpiImplementation`, defaults to `OpenMPI`
- The Intel implementation requires a Service fronting the launcher
- Passing the number of slots through environment variable instead of hostfile (some versions of Intel MPI ignore the slots defined in the hostfile).
- Add an example that uses Intel MPI

Intel MPI is very flaky at startup, as opposed to OpenMPI. In particular, it won't retry connections to workers if they are the hostnames are not resolvable. The entrypoint handles this.

### DIFF
--- a/examples/pi/intel-entrypoint.sh
+++ b/examples/pi/intel-entrypoint.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+set_intel_vars=/opt/intel/oneapi/setvars.sh
+if [ -f $set_intel_vars ]; then
+  source $set_intel_vars
+fi
+
+function resolve_host() {
+  host="$1"
+  check="nslookup $host"
+  max_retry=5
+  counter=0
+  backoff=0.1
+  until $check > /dev/null
+  do
+    if [ $counter -eq $max_retry ]; then
+      echo "Couldn't resolve $host"
+      return
+    fi
+    sleep $backoff
+    echo "Couldn't resolve $host... Retrying"
+    ((counter++))
+    backoff=$(echo - | awk "{print $backoff + $backoff}")
+  done
+  echo "Resolved $host"
+}
+
+if [ "$K_MPI_JOB_ROLE" == "launcher" ]; then
+  resolve_host "$HOSTNAME"
+  cat /etc/mpi/hostfile | while read host
+  do
+    resolve_host $host
+  done
+fi
+
+exec "$@"
--- a/v2/pkg/apis/kubeflow/v2beta1/types.go
+++ b/v2/pkg/apis/kubeflow/v2beta1/types.go
@@ -55,6 +55,10 @@ type MPIJobSpec struct {
 	// SSHAuthMountPath is the directory where SSH keys are mounted.
 	// Defaults to "/root/.ssh".
 	SSHAuthMountPath string `json:"sshAuthMountPath,omitempty"`
+
+	// MPIImplementation is the MPI implementation.
+	// Options are "OpenMPI" (default) and "Intel".
+	MPIImplementation MPIImplementation `json:"mpiImplementation,omitempty"`
 }
 
 // MPIReplicaType is the type for MPIReplica.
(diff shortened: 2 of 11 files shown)


terrytangyuan (commented):
Others who have experience please help review this one.

alculquicondor:
cc @kawych

gaocegege:
cc @zidarko

gaocegege (commented):
/cc @zw0610 @carmark

v2/pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -67,3 +71,10 @@ const (
 	// MPIReplicaTypeWorker is the type for worker replicas.
 	MPIReplicaTypeWorker MPIReplicaType = "Worker"
 )
+
+type MPIImplementation string
+
+const (
+	OpenMPIImplementation  MPIImplementation = "OpenMPI"
```
gaocegege, line 78: Can we follow the convention like MPIReplicaType:

```
type MPIImplementation string

const (
	MPIImplementationOpenMPI  MPIImplementation = "OpenMPI"
	MPIImplementationIntel          MPIImplementation = "Intel"
```

v2/pkg/controller/mpi_job_controller.go
```diff
@@ -711,25 +748,31 @@ func (c *MPIJobController) getOrCreateConfigMap(mpiJob *kubeflow.MPIJob, isGPULa
 
 // getOrCreateWorkerService gets the workers' Service controlled by this MPIJob,
 // or creates one if it doesn't exist.
-func (c *MPIJobController) getOrCreateWorkersService(mpiJob *kubeflow.MPIJob) (*corev1.Service, error) {
-	svc, err := c.serviceLister.Services(mpiJob.Namespace).Get(mpiJob.Name + workerSuffix)
+func (c *MPIJobController) getOrCreateWorkersService(job *kubeflow.MPIJob) (*corev1.Service, error) {
+	return c.getOrCreateService(job, newWorkersService)
```
kawych, line 752: Can you simplify the code by creating the service here and removing the newWorkersService() factory function? And same with the lanucher?

examples/pi/intel-entrypoint.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+set_intel_vars=/opt/intel/oneapi/setvars.sh
+if [ -f $set_intel_vars ]; then
+  source $set_intel_vars
+fi
+
+function resolve_host() {
```
kawych, line 8: I think this still can cause issues sometimes, i.e. sometimes there is a window when the launcher is able to resolve it's own hostname, but workers are not. This happens to me usually in the second run if I schedule two runs in a row.

It should be fine though, since this is only an example

kawych (approved):
LGTM. Would be great if you could document what Intel MPI versions was it tested with.

examples/pi/intel-entrypoint.sh
alculquicondor: Are you sure that was the problem?
When I only had the check for the launcher, I was getting flaky startups. Now that I have checks for every workers as well, the job starts every time. I couldn't really debug what was going on with just the check for the launcher, as Hydra  doesn't log the output of the ssh calls :(

v2/pkg/controller/mpi_job_controller.go
alculquicondor: I'm getting rid of the factory in the `getOrCreateService` function. But I have to keep the `newWorkersService` because it's used for tests.

alculquicondor:
Any other suggestions?

gaocegege:
LGTM :+1:
/lgtm

alculquicondor:
/assign @terrytangyuan
for approval

terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #390, opened 2021-07-28
Optimize OpenMPI image size

alculquicondor:
Removing the compiler from the final image.
From 423MB to 184MB


terrytangyuan (commented):
Thanks!

/lgtm

gaocegege (commented):
/lgtm

Multi-stage build is exactly what we need!

/assign @terrytangyuan

Issue #394, opened 2021-08-06
Launcher service account is always overwritten 


Dimss:
There is no way to use existing RBAC configurations (`ServiceAccount`, `Role` and `RoleBinding`) for `Launcher` containers. 
The MPIJob spec includes K8s pod template, where end user may configure desired `ServiceAccountName`
This configuration works as expected with `Worker` spec, but not with `Launcher` spec. 

For example, configuring `spec.mpiReplicaSpecs.Worker.template.spec.serviceAccountName=cnvrg` will cause all the workers pods to run with `cnvrg` service account.
However configuring `spec.mpiReplicaSpecs.Launcher.template.spec.serviceAccountName=cnvrg` **will not cause** all the launcher pods to turn with `cnvrg` service account, instead MPI Operator will  override this configuration, and will create a new `ServiceAccount`, `Role` and `RoleBinding` dedicated for the current MPIJob object, and the name of the objects will be `mpiJob.Name + launcherSuffix`

alculquicondor:
@Dimss consider trying the v2 controller, which doesn't depend on ServiceAccounts :)

Dimss:
@alculquicondor two questions 
1. is v2 controller is production grade? 
2. is v2 controller is serving same CRD spec as `v1alpha2`

alculquicondor:
1. I'm working towards it. So far it has more coverage that the previous versions. But no-one is using it in production AFAIK.
2. No, I'm taking the chance to work towards a consistent API. It would be ideal to have a conversion webhook, but I might not have time for it. I would appreciate if someone else can take that lead, though.

Dimss:
@alculquicondor thank you for your response. 
So, in that case, I'll have to continue to use `v1alpha2`, and hopefully my PR (#395) will be approved.

alculquicondor:
I'd appreciate if you still have a chance to try the v2 controller. You can follow along with the progress in #373

alculquicondor:
btw @Dimss: the v2 controller is now released as v0.3.0.

PR #395, opened 2021-08-06
Add support for using existing ServiceAccount for Launcher pod (#394)

Dimss:
Add support for using existing ServiceAccount for Launcher pod (#394)

* Do not create a SA, Role and RoleBinding when --use-launcher-pod-spec-serviceaccount=true,
  instead, use SA configured in spec.mpiReplicaSpecs.Launcher.template.spec.serviceAccountName

### DIFF
--- a/cmd/mpi-operator.v1alpha2/app/options/options.go
+++ b/cmd/mpi-operator.v1alpha2/app/options/options.go
@@ -23,15 +23,16 @@ import (
 
 // ServerOption is the main context object for the controller manager.
 type ServerOption struct {
-	Kubeconfig           string
-	MasterURL            string
-	KubectlDeliveryImage string
-	Threadiness          int
-	MonitoringPort       int
-	PrintVersion         bool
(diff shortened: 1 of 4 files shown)


terrytangyuan (commented):
This looks good but perhaps adding some error/warning messages when there are errors related to insufficient service accounts?

Dimss:
@terrytangyuan Hmm, I'm not sure I exactly understand what do you mean. 
Basically, if user set `--use-launcher-pod-spec-serviceaccount=true`, I'll first check if such a SA exists, if the SA doesn't exists, Mpi Operator log will throw an error, and the reconcile loop will never finished. 

So, when you saying "adding some error/warning messages" do you mean adding more logs prints? 
or updating the `spec.status` fields? 
or something else?

terrytangyuan:
Okay then we should be good. LGTM from me. Others PTAL.

terrytangyuan (approved):
/approve

PR #397, opened 2021-08-10
Fix Discovery script for intel

alculquicondor:
Follow up to #389

Slots are handled through an environment variable instead.


terrytangyuan (commented):
Do non-Intel implementation need this file?

alculquicondor:
AFAIK, yes. But this new structure should work for both implementations. We are passing number of slots through environment variables now.

terrytangyuan (approved):
/lgtm
/approve

PR #398, opened 2021-08-10
Remove ability to run ranks in launcher

### LINKED ISSUE #393: Reevaluate: launcher can also run the workload
Context:
- Making the launcher also run the workload was initially proposed in #197 and implemented in #276 and #363
- I'm trying to ensure that the API is well defined

Paraphrased from https://github.com/kubeflow/mpi-operator/issues/197#issuecomment-867040347:

In my opinion, having the workload run in the launcher is ambiguous. Sometimes you have N workers (where N is the number specified in the worker replicas) and sometimes you have N+1. This is decided by just having a request for a resource that has "gpu" as a substring, which might have collisions with other user-defined requests.

Another potential problem is that users don't have fine control of how much of the requests/limits should go to the launcher process and how much to the workload process running in the same container.

If the idea is to be able to run a worker in the same node as the launcher, this can still be achieved by scheduling 2 pods on it (one for the worker and one for the launcher), setting the requests appropriately. Such behavior is more explicit from an API perspective. This setup also allows users who don't use GPUs to a workload in the same node as the launcher.

Alternatively, we could have an explicit field `.spec.runWorkloadInLauncher` or `.spec.LauncherPolicy`. But I believe good documentation around proper requests should be enough. Administrators could enforce that launchers don't requests GPUs in a webhook.

alculquicondor:
Remove the logic to detect if a Pod requests GPU to decide to include the launcher in the hostfile.

Fixes #393

### DIFF
--- a/v2/pkg/controller/mpi_job_controller.go
+++ b/v2/pkg/controller/mpi_job_controller.go
@@ -26,7 +26,6 @@ import (
 	"reflect"
 	"sort"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


gaocegege (commented):
/assign @carmark @zw0610

terrytangyuan (approved):
/approve

alculquicondor:
I rebased and need new lgtm.

alculquicondor:
@terrytangyuan PTAL

Issue #296, opened 2020-10-22
Add e2e workflows against AWS CI


Jeffwan:
Community is asking different WG to own their infra and community won't provide a common shared testing infra anymore. Sees kubeflow/testing#752 for more details.

Pytorch migration works well and here's the PR to update test case to work on AWS Prow+ Argo cluster.
kubeflow/pytorch-operator#305

Seems mpi-operator has empty workflow defined https://github.com/kubeflow/mpi-operator/blob/master/prow_config.yaml
Whenever it plans to add real e2e tests, it can follow what pytorch operator did and test each PR before the merge.

terrytangyuan:
cc @kubeflow/wg-training-leads

alculquicondor:
@Jeffwan, I would like to add E2E tests.

> Community is asking different WG to own their infra and community won't provide a common shared testing infra anymore.

Does this mean that the mpi-operator project has to find where to run their tests? Or is there any common infra already in place for the training WG? If so, Is there any guidebook to start adding the workflows and tests?

Jeffwan:
> @Jeffwan, I would like to add E2E tests.
> 
> > Community is asking different WG to own their infra and community won't provide a common shared testing infra anymore.
> 
> Does this mean that the mpi-operator project has to find where to run their tests? Or is there any common infra already in place for the training WG? If so, Is there any guidebook to start adding the workflows and tests?

https://github.com/kubeflow/testing/tree/master/aws We use this infra for e2e testing. It can provision the cluster and you can run test against it.

alculquicondor:
We added local tests using kind. This should be enough for now. But thanks for the pointer, maybe it will be useful in the future for more complex workloads.

PR #399, opened 2021-08-12
E2E test for v2 controller

alculquicondor:
Part of #373. Related to #296 

Add first E2E test for the v2 controller: Tests that an OpenMPI sample runs to completion as non-root user.

The tests run locally on a [kind](https://kind.sigs.k8s.io/docs/user/quick-start/) cluster.

The tests run as a new workflow job.

**Notes to reviewer**

Other tests combinations (IntelMPI, run-as-root) coming in follow up PRs.

### DIFF
--- a/v2/test/e2e/e2e_suite_test.go
+++ b/v2/test/e2e/e2e_suite_test.go
@@ -1,0 +1,157 @@
+// Copyright 2021 The Kubeflow Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"testing"
+	"time"
+
+	clientset "github.com/kubeflow/mpi-operator/v2/pkg/client/clientset/versioned"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	envUseExistingCluster   = "USE_EXISTING_CLUSTER"
+	envTestMPIOperatorImage = "TEST_MPI_OPERATOR_IMAGE"
+	envTestKindImage        = "TEST_KIND_IMAGE"
+
+	defaultMPIOperatorImage = "kubeflow/mpi-operator:local"
+	defaultKindImage        = "kindest/node:v1.21.2"
+	openMPIImage            = "kubeflow/mpi-pi:openmpi"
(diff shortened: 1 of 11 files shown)


alculquicondor:
Ready for review

gaocegege (commented):

v2/test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	envTestUseExistingCluster = "TEST_USE_EXISTING_CLUSTER"
```
kawych, line 37: nit: Does this follow a convention for other tests? If not, I suggest to have something like "TEST_CREATE_CLUSTER" (negation of this).

v2/test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"testing"
+	"time"
+
+	clientset "github.com/kubeflow/mpi-operator/v2/pkg/client/clientset/versioned"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
```
kawych, line 28: nit: dot import (. "github.com/onsi/gomega") will allow you shorten some code: "Expect(err).ToNot(HaveOccurred())"

v2/test/e2e/mpi_job_test.go
```diff
@@ -0,0 +1,159 @@
...
+			},
+		}
+		var err error
+		ns, err = k8sClient.CoreV1().Namespaces().Create(context.Background(), ns, metav1.CreateOptions{})
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		namespace = ns.Name
+	})
+
+	ginkgo.AfterEach(func() {
+		if namespace != "" {
```
kawych, line 50: Can it be empty?

v2/test/e2e/mpi_job_test.go
```diff
@@ -0,0 +1,159 @@
...
+			err := k8sClient.CoreV1().Namespaces().Delete(context.Background(), namespace, metav1.DeleteOptions{})
+			if !errors.IsNotFound(err) {
+				gomega.Expect(err).ToNot(gomega.HaveOccurred())
+			}
+		}
+	})
+
+	ginkgo.BeforeEach(func() {
```
kawych, line 58: It's maybe more philosophical, but I think that all logic specific to openMPI should be encapsulated within "ginkgo.Context("with OpenMPI implementation", func() { ... }). Is the mpiJob's spec identical for intelMPI apart from the image? Would it work to have a function "func getPiMPIJob(image string)" and use it as "getPiMPIJob(openMPIImage)"? If not, maybe just move it to the OpenMPI context?

alculquicondor: The command changes a little too. Moved to Context

v2/test/e2e/mpi_job_test.go
kawych: I see, so "Expect(err).ToNot(HaveOccurred())" doesn't stop the test from executing?

Then I think it's worth to break the test execution when it clearly doesn't make sense anymore, for example when the MPI Job's creation fails. Even when namespace creation fails, continuation gives little hope for any useful information.

alculquicondor: It does stop the test from executing, but `AfterEach` will still be called.

kawych (approved):

gaocegege (commented):
/assign @terrytangyuan @zw0610

terrytangyuan (approved):
This is great. Thanks!

/lgtm
/approve

PR #409, opened 2021-08-19
Use fully-qualified label names from common

alculquicondor:
Part of #373

This is a breaking change with respect to the v1 controller (if users are querying pods with the previous labels).

If this is not desired, we can add both new and deprecated labels, but I think this is acceptable given that a lot of the v2 controller breaks backwards compatibility.


gaocegege (commented):
/lgtm

terrytangyuan (approved):
/approve

Issue #386, opened 2021-07-27
 Allow launcher to start after workers are ready


alculquicondor:
Part of #373. This was initially discussed in #360, but it didn't have a final conclusion.

In the v1 controller, an init-container was responsible of holding the start of the launcher until all the workers were ready.
In the v2 controller, we removed this init container for scalability and performance reasons (each init container is a controller that has to build a full cache of all the pods in the cluster).

Most of the times, this just works. The workers and launcher start at roughly the same time. ssh itself does some retries before giving up. However, there are scenarios were the launcher fails because it can't find the workers.
This problem could be consistent in the following scenario: The launcher lands in a Node where the image is already present, but the workers launch in Nodes where the image is missing.

There are 2 high level solutions to this problem:
1. Delay the creation of the launcher Pod. This is possible by watching the Running pods in the controller. The problem here is that we might delay the start up of a Job significantly (in some scenarios, the user has to wait for 2 image pulls to happen sequentially, instead of all the pulls to happen significantly).
2. Retry the launcher. We can get this "for-free" if we use a Job for the launcher. The Job controller has exponential backoff for failures. The number of retries is configurable.

Both solutions are not exclusive (we can implement both).
Additional questions come with both solutions:

1. Should we make the creation delay an API option?
2. Should we expose the number of retries to the MPIJob user?

I propose we do 2 first, exposing the number of retries in the API, and then re-evaluate if we need to use 1 from user feedback after the release of a 2.0.

alculquicondor:
cc @gaocegege @terrytangyuan @ahg-g @kawych

terrytangyuan:
Agreed. The second option looks better and could leverage a Job.

ahg-g:
SGTM. +1 to using the job api wherever possible.

xhejtman:
the second is functional, but looks a bit nasty to me:
![mpi-launcher](https://user-images.githubusercontent.com/22795052/130798770-b7b63c99-b51f-4e04-b983-28e691cb7b5d.png)

xhejtman:
but may be it solvable just to use another type of deployment of launcher job so that it restarts in place?

alculquicondor:
You can definitely use `restartPolicy: OnFailure`

xhejtman:
could it be a default then?

alculquicondor:
The default in k8s is `Always`, which is not allowed for Jobs. I guess having `OnFailure` for MPIJob is fair. Do you mind sending a PR for that?

[...]

PR #416, opened 2021-08-24
Mount SSH Secret directly on main container

### LINKED ISSUE #407: Wrong network rules for horovod example
Hello,

I use kubeflow 1.3.0, mpi-operator latest from Jul 5 2021. I tried MPI Job example horovod, however, launcher deployment has error in kubectl-delivery pod:
Failed to list *v1.Pod: Get https://10.43.0.1:443/api/v1/namespaces/adrian/pods?limit=500&resourceVersion=0: dial tcp 10.43.0.1:443: connect: connection refused

I tested, the pod is unable to reach any outer node or external internet on ports 80 or 443. I tested those two workers, they are OK, they can reach anything, one of the workers is running on the same physical node as launcher and worker can reach network, launcher cannot. Launcher can ping interneter (e.g. 8.8.8.8). 

Is it some istio misconfiguration problem? I tried to forcibly start launcher on a different node, result is the same. Thanks for help.

alculquicondor:
Remove the init container for faster startup.

Possible by disabling StrictModes in sshd_config

Fixes #407


alculquicondor:
@xhejtman can you please take a look?

alculquicondor:
The E2E for Intel is failing. However, it doesn't seem to be due to this change. The existing test was actually running the OpenMPI image by mistake.

The test success locally, but not in github action. I'm trying to debug the reason.

xhejtman:
> @xhejtman can you please take a look?

Just tried it, it seems to be ok. However, there is still minor issue that launcher creates 3 error pods and 4th that is running. This is beacuse workers are not DNS resolvable. Perhaps we still need init container that will wait for DNS to populate names?

alculquicondor:
We discussed that in #386

PR #420, opened 2021-08-30
Set OnFailure default restart policy for launcher

xhejtman:
Change Launger restart policy from Newer to OnFailure to mitigate initial DNS lookup errors.

### DIFF
--- a/v2/pkg/apis/kubeflow/v2beta1/constants.go
+++ b/v2/pkg/apis/kubeflow/v2beta1/constants.go
@@ -21,6 +21,8 @@ const (
 	EnvKubeflowNamespace = "KUBEFLOW_NAMESPACE"
 	// DefaultRestartPolicy is default RestartPolicy for ReplicaSpec.
 	DefaultRestartPolicy = common.RestartPolicyNever
+	// Launcher will fail on start if workers DNS names are not ready so we set OnFailure restart policy
+	DefaultLauncherRestartPolicy = common.RestartPolicyOnFailure
 	// OperatorName is the name of the operator used as value to the label common.OperatorLabelName
 	OperatorName = "mpi-operator"
 )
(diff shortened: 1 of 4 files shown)


xhejtman:
@googlebot I signed it!

terrytangyuan (commented):
Should this be specified by the user instead of being as default in the controller?

xhejtman:
User can specify whatever he wants, I believe OnFailure is good default untill better solution is found here: https://github.com/kubeflow/mpi-operator/issues/386

If Never is the default, you will see several failed pods for launcher.

terrytangyuan (approved):
/lgtm
/approve

v2/pkg/apis/kubeflow/v2beta1/constants.go
```diff
@@ -21,6 +21,8 @@ const (
 	EnvKubeflowNamespace = "KUBEFLOW_NAMESPACE"
 	// DefaultRestartPolicy is default RestartPolicy for ReplicaSpec.
 	DefaultRestartPolicy = common.RestartPolicyNever
+	// Launcher will fail on start if workers DNS names are not ready so we set OnFailure restart policy
```
alculquicondor, line 24: This comment is part of the godoc. I don't think we need to state the reasons for the change here. Just document what this is.

v2/pkg/apis/kubeflow/v2beta1/default.go
```diff
@@ -29,7 +29,7 @@ func setDefaultsTypeLauncher(spec *common.ReplicaSpec) {
 		return
 	}
 	if spec.RestartPolicy == "" {
-		spec.RestartPolicy = DefaultRestartPolicy
+		spec.RestartPolicy = DefaultLauncherRestartPolicy
```
alculquicondor, line 32: We have unit tests for this, so fix them :)

xhejtman:
Hopefully, it is fixed.

PR #425, opened 2021-09-08
Add readiness probe to Intel MPI jobs

alculquicondor:
This improves the reliability of MPI Jobs.

The readiness probe ensures that sshd is up and running before the hostname is resolvable.


terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #428, opened 2021-09-13
Set ClusterFirstWithHostNet DNS policy when the Pods use host network.

alculquicondor:
This allows resolving the worker and launcher hostnames without needing to include the namespace or cluster domain.

Also migrated the base image to use port 2222, as port 22 is usually taken by the host's sshd. This should make it easier for pods to use host network.


gaocegege (commented):
LGTM

BTW, can you please explain the use scenario of using host network?

Is it for performance issue or do you want to use RDMA?

alculquicondor:
Some users might choose host network for extra network performance.

PR #449, opened 2022-01-17
Support read permissions for clusterRole and write permissions for mu…

xiaonancc77:
In a multi-user shared kubernetes cluster, it is difficult to provide cluster-level write permissions to mpi-operator, but it is necessary for mpi-operator to dynamically support the ability to create mpijobs in multiple namespaces.
Therefore, mpi operaetor needs to support cluster-level read permissions (get/list/watch) and multiple namespace-level write permissions.
When adding a new namespace, you need to create mpijobs, you can create Role and Rolebindng in the corresponding namespace.

### DIFF
--- a/cmd/mpi-operator.v1/app/options/options.go
+++ b/cmd/mpi-operator.v1/app/options/options.go
@@ -24,6 +24,7 @@ import (
 // ServerOption is the main context object for the controller manager.
 type ServerOption struct {
 	Kubeconfig           string
+	KubeconfigRorWrite   string
 	MasterURL            string
 	KubectlDeliveryImage string
 	Threadiness          int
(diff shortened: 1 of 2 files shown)


alculquicondor:
I'm failing to see why you need 2 kubeconfigs. Can you elaborate?

xiaonancc77:
> I'm failing to see why 2 kubeconfigs would solve the problem. Can you elaborate?
For security reasons, a multi-user shared cluster cannot provide an operator with cluster-level update create permissions, but it is expected that this operator can dynamically create MPIJobs for multiple namespaces. Users can apply for the create permission of a namespace from O&M. 
An Informer can watch list multiple namespace resources only after cluster-level permissions are configured. So a KubeconFig provides cluster-level watch list get permissions for informer and a KubeconFig provides namespace level create update permissions for client. You can specify a namespace to create resources. Therefore, dynamic adjustment can support the creation of namespace of resources through the user's RABC.

alculquicondor:
I feel like you should be able to achieve such behavior with a single kubeconfig by tweaking the cluster roles https://github.com/kubeflow/mpi-operator/blob/master/manifests/base/cluster-role.yaml

Is the pattern you are proposing used somewhere else?

xiaonancc77:
> I feel like you should be able to achieve such behavior with a single kubeconfig by tweaking the cluster roles https://github.com/kubeflow/mpi-operator/blob/master/manifests/base/cluster-role.yaml
> 
> Is the pattern you are proposing used somewhere else?


Thanks, I tried to bind the clusterRole and namespace scope  roles with one user, and it works the same. I will close this pull request.

chunfengzhao1:
@xiaonancc77 could give details on how to bind clusterRole and namespace scope roles to make multi-namespace work? one example should help. Thanks!

PR #452, opened 2022-02-07
Build CRDs using kubebuilder

### LINKED ISSUE #408: Build CRDs using kubebuilder
Currently, CRDs are written manually. kubebuilder should be able to handle this for us, while guaranteeing that the API remains structured under the OpenAPI v3 definition.

ArangoGutierrez:
Fixes #408

Signed-off-by: Carlos Eduardo Arango Gutierrez <carangog@redhat.com>


v2/crd/kubeflow.org_mpijobs.yaml
```diff
@@ -0,0 +1,6908 @@
+
```
alculquicondor: Shouldn't this replace the old file?

ArangoGutierrez: not yet, I want to also patch the commons api so the generated CRD has defaults as it has right now

ArangoGutierrez: For now, this patch just demonstrates that the CRD can be generated, and shows how it looks

alculquicondor: gotcha. Let's get the changes in the common api first.

terrytangyuan (commented):
kubeflow/training-operator already uses kubebuilder. Perhaps send a PR there instead? And then v2 controller can switch to use kubebuilder as well when it is migrated to that repo.

alculquicondor:
Should we proceed with this @terrytangyuan?

terrytangyuan:
For v2 yes.

v2/pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -40,36 +40,37 @@ type MPIJobList struct {
...
-const (
-	// MPIReplicaTypeLauncher is the type for launcher replica.
-	MPIReplicaTypeLauncher MPIReplicaType = "Launcher"
-
-	// MPIReplicaTypeWorker is the type for worker replicas.
-	MPIReplicaTypeWorker MPIReplicaType = "Worker"
-)
+type MPIReplicaSpecs struct {
+	// +kubebuilder:default:={Replicas:1,RestartPolicy:"OnFailure"}
+	Launcher *common.ReplicaSpec `json:"launcher"`
```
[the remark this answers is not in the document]
ArangoGutierrez: yeah rolled back on this change, it will be too big, a lot of tests cases break on this. will have it as a TODO if we move to k8s-sigs

ArangoGutierrez:
Ok I think this is as far as I can go using kubebuilder here 
@alculquicondor @terrytangyuan Ready for review again

terrytangyuan (approved):
Thanks!

/lgtm
/approve

Issue #445, opened 2021-12-07
Elastic resize broken in v2 operator: Horovod requires the slots parameter


alculquicondor:
In the v2 controller, the slots parameter in the hostfile was replaced by environment variables:

https://github.com/kubeflow/mpi-operator/blob/b88edad03a4b55708deaed5f8ebe058eeacf19bd/v2/pkg/controller/mpi_job_controller.go#L116-L117

However, horovod still requires the number of slots. So we need to put the parameter back. The duplication of information should be fine.
It's unclear if horovod supports intel.

We need an E2E test to ensure that the fix works long term.

alculquicondor:
Also, we need to adapt the intel entry point to the change of the hostfile format: https://github.com/kubeflow/mpi-operator/blob/master/examples/base/intel-entrypoint.sh

alculquicondor:
@tenzen-y we should fix this as part of the next release #507, but I don't have bandwidth. Do you?

tenzen-y:
> @tenzen-y we should fix this as part of the next release #507, but I don't have bandwidth. Do you?

@alculquicondor Thanks for mentioning this. Sure.

IIUC, we need to modify the hostfile in https://github.com/kubeflow/mpi-operator/blob/31d4575066770756cc8ca347cc6d5ba3113b1e02/pkg/controller/mpi_job_controller.go#L1186
to `buffer.WriteString(fmt.Sprintf("%s%s-%d.%s slots=%d\n", mpiJob.Name, workerSuffix, i, workersService, slots))`, right? 

Also, we need to adapt the new hostfile format in https://github.com/kubeflow/mpi-operator/blob/db6930dcd509b1d97a47bcccd1bcb1da5eaea712/examples/base/intel-entrypoint.sh#L30.

PR #478, opened 2022-10-11
MPICH support

### LINKED ISSUE #477: MPICH question
I have a Docker image of my python program which calls into parallel version of meep (https://meep.readthedocs.io/en/latest/) which is compiled against the MPICH (pymeep package in conda-forge). I'd like to run it on the K8s using MPI-Operator. I see 3 possible paths:

1. Extend MPI-Operator to support MPICH
2. Recompile Parallel Meep to use OpenMPI
3. This whole things won't work either way because of shenanigans of havin MPI enabled app inside conda environemnt in docker which is not supported.

Any advice which is better? Will it even work or am I doomed and I'll hit point 3 above?

sheevy:
This is my initial attempt to have MPICH working on mpi-operator. Bear in mind I'm a noob when it comes to MPI and Kubernetes.

This is to close #477 which @alculquicondor offered to help with. 

I've demonstrated pi example works when using MPICH.

There's a single TODO in the commit, because I'm not sure what a value of one of the constant should be for MPICH.

I've changed some of the test files, but not all. I'd appreciate a hint how to progress on that front.

Also, the autogenerated files were in fact edited by me manually...

### DIFF
--- a/v2/pkg/controller/mpi_job_controller.go
+++ b/v2/pkg/controller/mpi_job_controller.go
@@ -115,6 +115,7 @@ const (
 
 	openMPISlotsEnv  = "OMPI_MCA_orte_set_default_slots"
 	intelMPISlotsEnv = "I_MPI_PERHOST"
+	mpichSlotsEnv    = "TODO"
 )
 
 var (
@@ -196,6 +197,16 @@ var (
 		},
 		{
 			Name:  "I_MPI_HYDRA_BOOTSTRAP_EXEC_EXTRA_ARGS",
+			Value: "-o ConnectionAttempts=10",
+		},
+	}
+	mpichEnvVars = []corev1.EnvVar{
+		{
+			Name:  "HYDRA_HOST_FILE",
+			Value: fmt.Sprintf("%s/%s", configMountPath, hostfileName),
+		},
+		{
+			Name:  "HYDRA_LAUNCH_EXTRA_ARGS",
 			Value: "-o ConnectionAttempts=10",
 		},
 	}
@@ -560,8 +571,9 @@ func (c *MPIJobController) syncHandler(key string) error {
 		if err != nil {
 			return err
 		}
-		if mpiJob.Spec.MPIImplementation == kubeflow.MPIImplementationIntel {
-			// The Intel implementation requires workers to communicate with the
+		if mpiJob.Spec.MPIImplementation == kubeflow.MPIImplementationIntel ||
+		   mpiJob.Spec.MPIImplementation == kubeflow.MPIImplementationMPICH {
+			// The Intel and MPICH implementations require workers to communicate with the
 			// launcher through its hostname. For that, we create a Service which
 			// has the same name as the launcher's hostname.
 			_, err := c.getOrCreateService(mpiJob, newLauncherService(mpiJob))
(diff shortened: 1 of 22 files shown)


v2/pkg/controller/mpi_job_controller.go
```diff
@@ -115,6 +115,7 @@ const (
 
 	openMPISlotsEnv  = "OMPI_MCA_orte_set_default_slots"
 	intelMPISlotsEnv = "I_MPI_PERHOST"
+	mpichSlotsEnv    = "TODO"
```
sheevy, line 118: Does anyone know how to figure out what should be provided here?
Interestingly, the pi example works as is, i.e. with mpichSlotsEnv == TODO

alculquicondor: This should be the name of the environment variable for configuring the number of workers per host.

It works in the sense that it runs, because this would just set an environment variable called TODO. But it's not doing the intended outcome.

alculquicondor: In general, it's hard to overwrite the command line, because users might be using wrappers around `mpiexec`. That's why environment variables is better. Another option that openMPI and Intel MPI support is to set them in the hostfile.

Actually, it might be useful to use that in general because of this #445

alculquicondor:
Thanks, this PR is still in my radar, but I don't think I'll have a chance to look at it in the next 2 weeks.

sheevy:
Thanks. Are there any actions on me at this point?

alculquicondor:
Did you figure a solution for this?

```
mpichSlotsEnv    = "TODO"
```

sheevy:
Not yet. I think I misunderstood the purpose od `-ppn` option. Just asked a follow up question in MPICH repo: https://github.com/pmodels/mpich/discussions/6209

sheevy:
Thanks for the discussion in the other thread. So, with my limit understanding, the next step for me would be to remove the part with sets up `mpichSlotsEnv` env variable, and use `SlotsPerWorker` from the spec to drive the hostfile generation. And it'll be on the user to know to skip the `-n` argument.
Is that correct understanding?

alculquicondor:
That is correct. It's up to the user to set `-n` if they want to, but they should be able to skip it.

alculquicondor:
And if you have a chance, maybe we could be consistent across the MPI implementations and always use the hostfile instead of an environment variable.

That would fix https://github.com/kubeflow/mpi-operator/issues/445 too

alculquicondor:
@sheevy are you still working on this?

alculquicondor:
can you rebase?

sheevy:
oopsie, sorry for noise, this was meant to to be a local change in my fork. I forgot the PR is still open.

I started reappling the change from scratch onto master as rebasing this branch was just too much work. Too much has changed in the time this PR was dormant. But I now have time to finalyl close it.

As I understand, the recent addtion of slots to host file is exactly what was missing. So now it's just a case of porting the change from this PR onto master ,as everything else is now in place.

tenzen-y:
@sheevy Do we yet need this PR? If not, we can close this PR in favor of #562.

alculquicondor:
/close
I suspect not

Issue #504, opened 2023-01-24
Add suspend semantics


alculquicondor:
The semantics should be similar to that of k8s Job.

And this will pave the work for the training-operator (https://github.com/kubeflow/training-operator/issues/1519)

mimowo:
@tenzen-y @alculquicondor you may want to look at the WIP implementation (tested manually) here: https://github.com/kubeflow/mpi-operator/pull/511. Any early feedback is welcome.

tenzen-y:
Probably, we can close this issue.

PR #510, opened 2023-01-27
Clean up manifests

tenzen-y:
Signed-off-by: Yuki Iwai <yuki.iwai.tz@gmail.com>

I cleaned up manifests in the following:

1. Replace the image tag with 'master' to avoid [this](https://github.com/kubeflow/mpi-operator/issues/508) confusing.
2. Update the `make crd` target so that we can add the latest CRD to the manifests directory.
3. Regenerate `deploy/v2beta1/mpi-operator.yaml`.
4. Update the Readme.md.
5. Add CI to verify generated codes.

---
Our operator images at dockerhub:

- `latest`: https://hub.docker.com/layers/mpioperator/mpi-operator/latest/images/sha256-3ccfa8d8b7bf97836b291f5cadb39afec4b3612acddb4a5f81ecccdfaede92f9?context=explore

- `master`: https://hub.docker.com/layers/mpioperator/mpi-operator/master/images/sha256-84749d84196f15fa61b2a2786ae625b4b85dd750dd4dd5e39970d4985ebfc70a?context=explore

Fixes: #508

PR #511, opened 2023-01-27
WIP: Support suspend semantics for MPIJob

mimowo:
It solves: https://github.com/kubeflow/mpi-operator/issues/504

WIP because:
- no tests yet
- it places the `Suspend` field under Spec, but it seems it should be under `RunPolicy` so that it is in commons, similarly as in this PR: https://github.com/kubeflow/common/pull/193
- there is yet no `JobSuspended` condition, but it does not seem necessary. Though, again the condition types are in commons, so it could be 2 additional PRs: introduce conditions in the commons and use them here.


mimowo:
@alculquicondor @tenzen-y WIP but ready for early feedback (would be good as this is my first PR in this repo). PTAL.

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -62,6 +62,18 @@ type MPIJobSpec struct {
 	// +kubebuilder:validation:Enum:=OpenMPI;Intel
 	// +kubebuilder:default:=OpenMPI
 	MPIImplementation MPIImplementation `json:"mpiImplementation,omitempty"`
+
+	// suspend specifies whether the Job controller should create Pods or not. If
```
alculquicondor, line 66: s/Job/MPIJob

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -62,6 +62,18 @@ type MPIJobSpec struct {
...
+	// a Job is created with suspend set to true, no Pods are created by the Job
+	// controller. If a Job is suspended after creation (i.e. the flag goes from
+	// false to true), the Job controller will delete all active Pods associated
+	// with this Job. Users must design their workload to gracefully handle this.
+	// Suspending a Job will reset the StartTime field of the Job, effectively
+	// resetting the ActiveDeadlineSeconds timer too. Defaults to false.
```
alculquicondor, line 72:
```suggestion
	// resetting the activeDeadlineSeconds timer too. Defaults to false.
```

pkg/controller/mpi_job_controller.go
```diff
@@ -585,9 +577,48 @@ func (c *MPIJobController) syncHandler(key string) error {
...
+	// cleanup the running worker pods if the MPI job is suspended
+	if isMPIJobSuspended(mpiJob) {
+		if err := cleanUpPods(mpiJob, c); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
+func cleanUpPods(mpiJob *kubeflow.MPIJob, c *MPIJobController) error {
```
alculquicondor, line 607:
```suggestion
func cleanUpWorkerPods(mpiJob *kubeflow.MPIJob, c *MPIJobController) error {
```

pkg/controller/mpi_job_controller.go
```diff
@@ -585,9 +577,48 @@ func (c *MPIJobController) syncHandler(key string) error {
 		return err
 	}
 
+	if launcher != nil {
+		launcherSuspendUpdate := false
+		if isMPIJobSuspended(mpiJob) && !isJobSuspended(launcher) {
```
alculquicondor, line 582: You could have a single condition:
```suggestion
		if isMPIJobSuspended(mpiJob) != isJobSuspended(launcher) {
```

pkg/controller/mpi_job_controller.go
```diff
@@ -585,9 +577,48 @@ func (c *MPIJobController) syncHandler(key string) error {
...
+	// set worker StatefulSet Replicas to 0.
+	if err := c.deleteWorkerPods(mpiJob); err != nil {
+		return err
+	}
+	initializeMPIJobStatuses(mpiJob, kubeflow.MPIReplicaTypeWorker)
+	if c.gangSchedulerName != "" {
+		if err := c.deletePodGroups(mpiJob); err != nil {
+			return err
+		}
+	}
```
alculquicondor, line 617: @tenzen-y are you familiar with the volcano integration?

I wonder if we need to remove the pod group on suspension. Does it matter?

pkg/controller/mpi_job_controller.go
```diff
@@ -555,10 +546,11 @@ func (c *MPIJobController) syncHandler(key string) error {
 				return err
 			}
 		}
-
-		worker, err = c.getOrCreateWorker(mpiJob)
-		if err != nil {
-			return err
+		if !isMPIJobSuspended(mpiJob) {
+			worker, err = c.getOrCreateWorker(mpiJob)
```
alculquicondor, line 550: also create the podgroup conditionally?

pkg/controller/mpi_job_controller.go
```diff
@@ -503,24 +505,13 @@ func (c *MPIJobController) syncHandler(key string) error {
...
-			mpiJob.Status.ReplicaStatuses[common.ReplicaType(kubeflow.MPIReplicaTypeWorker)].Active = 0
-			return c.updateStatusHandler(mpiJob)
+			return cleanUpPods(mpiJob, c)
 		}
 		return nil
 	}
 
 	// first set StartTime.
-	if mpiJob.Status.StartTime == nil {
+	if mpiJob.Status.StartTime == nil && !isMPIJobSuspended(mpiJob) {
```
alculquicondor, line 514: if the job was suspended, we should reset the StartTime. Double check how we do it in the job controller.

alculquicondor (commented):
FYI, the common repo is on its way to disappearing. I would suggest copying the RunPolicy struct here and adding the field.

mimowo:
> FYI, the common repo is on its way to disappearing. I would suggest copying the RunPolicy struct here and adding the field.

I see, but is common reused by other subprojects too, right? So we would also need to copy the contents of common into these repos. Sounds like a lot of work, maybe simple, but the diffs will be big and one needs to be careful, so not sure we want to block the suspend work on that? Also, is this effort already planned, or in progress @alculquicondor @tenzen-y ?

tenzen-y:
> > FYI, the common repo is on its way to disappearing. I would suggest copying the RunPolicy struct here and adding the field.
> 
> I see, but is common reused by other subprojects too, right? So we would also need to copy the contents of common into these repos. Sounds like a lot of work, maybe simple, but the diffs will be big and one needs to be careful, so not sure we want to block the suspend work on that? Also, is this effort already planned, or in progress @alculquicondor @tenzen-y ?

Yes, that's right. We are using common repo in training-operator. However, we are planning to consolidate common codes to the training-operator repo.

https://github.com/kubeflow/training-operator/issues/1714

tenzen-y:
> FYI, the common repo is on its way to disappearing. I would suggest copying the RunPolicy struct here and adding the field.

@alculquicondor I agree with adding a `suspend` member to `Runpolicy`. Although can we copy `RunPolicy` in a separate PR? Since I think copying the `Runpolicy` to this repo is another context with this PR.

mimowo:
> @alculquicondor I agree with adding a `suspend` member to `Runpolicy`. Although can we copy `RunPolicy` in a separate PR? Since I think copying the `Runpolicy` to this repo is another context with this PR.

What about the other constants, like the once defining conditions? I guess we could have a PR to just copy `RunPolicy` to mpi-operator to unblock this work, but keep the dependency on common@0.4.6 for the condition constants. Then, we can extend the set of MPIJob conditions by JobSuspended just in the mpi-operator. If this sounds good I can open a preparatory PR just to copy RunPolicy.

pkg/controller/mpi_job_controller.go
mimowo: Done

pkg/controller/mpi_job_controller.go
mimowo: Done

pkg/apis/kubeflow/v2beta1/types.go
mimowo: Done

pkg/apis/kubeflow/v2beta1/types.go
mimowo: Done

tenzen-y:
> > @alculquicondor I agree with adding a `suspend` member to `Runpolicy`. Although can we copy `RunPolicy` in a separate PR? Since I think copying the `Runpolicy` to this repo is another context with this PR.
> 
> What about the other constants, like the once defining conditions? I guess we could have a PR to just copy `RunPolicy` to mpi-operator to unblock this work, but keep the dependency on common@0.4.6 for the condition constants. Then, we can extend the set of MPIJob conditions by JobSuspended just in the mpi-operator. If this sounds good I can open a preparatory PR just to copy RunPolicy.

Sounds good to me. Although, let me know what other members think.

cc @alculquicondor @terrytangyuan

alculquicondor:
sgtm

pkg/controller/mpi_job_controller.go
alculquicondor: AFAIK, a podgroup is just a declaration that some pods should be treated as a unit. But a podgroup doesn't create other objects.

tenzen-y: The `scheduling.volcano.sh/v1beta1` PodGroups also have queueing logic. So we might to need to delete PodGroup to re-queue Pods.

tenzen-y: Maybe, we should create E2E with the volcano in a separate PR.

pkg/controller/mpi_job_controller.go
```diff
@@ -449,6 +450,7 @@ func (c *MPIJobController) processNextWorkItem() bool {
 // converge the two. It then updates the Status block of the MPIJob resource
 // with the current status of the resource.
 func (c *MPIJobController) syncHandler(key string) error {
+	klog.Infof("___ MYDEBUG starting for %s", key)
```
tenzen-y, line 453: Do we need this debug log?

pkg/controller/mpi_job_controller.go
alculquicondor: probably someone with volcano experience should do it.

terrytangyuan:
Sounds good

mimowo:
@tenzen-y @alculquicondor I've opened the preparatory PR here: https://github.com/kubeflow/mpi-operator/pull/513. Please review.

pkg/controller/mpi_job_controller.go
mimowo: Node, cleaned :)

mimowo:
@terrytangyuan Please approve CI

pkg/controller/mpi_job_controller.go
mimowo: Done - I've added the JobSuspended condition and replicated the semantics around StartTime. 

However, the `.spec.runPolicy.activeDeadlineSeconds` is actually respected for MPIJobs via batch.Job: https://github.com/kubeflow/mpi-operator/blob/382da780a7a9a23e6c9fd9dc8c93ba77aba81c48/pkg/controller/mpi_job_controller.go#L1320. 

This means that the changes aren't strictly required to enforce the timeout. WDYT?

pkg/controller/mpi_job_controller.go
mimowo: Now I create all resources for as long as the MPIJob is suspended. This is the common workflow for Kueue, where the Job is created suspended. In some cases, when the job never get unsuspended (for whatever reason), we can save on creating the objects.

mimowo:
@alculquicondor @tenzen-y I fixed the issues reported so far and added tests (integration and e2e), so moving it out of WIP status. Please review.

pkg/controller/mpi_job_controller.go
mimowo: For now reverted the change - sorry for back end forth - this is just to minimize the diff as the unit tests currently verify the objects are created. Let me know what you think.

pkg/controller/mpi_job_controller.go
mimowo: I looked at the volcano code a little bit on my own, but could not conclude if it may create the pods under some scenarios.

Thus, for now, I delete the PodGroups either, to be on the safe side. WDYT?

+1 for the idea of an e2e test employing with podgroups. I guess we could create a follow-up Issue and ask for people willing to do it.

tenzen-y: > Thus, for now, I delete the PodGroups either, to be on the safe side. WDYT?

That sounds good to me.

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -130,3 +142,8 @@ const (
 	MPIImplementationOpenMPI MPIImplementation = "OpenMPI"
 	MPIImplementationIntel   MPIImplementation = "Intel"
 )
+
+const (
+	// JobSuspended means the job has been suspended.
+	JobSuspended common.JobConditionType = "Suspended"
+)
```
tenzen-y, line 149: Maybe we want to move `common.JobConditionType` to this repo, although we can follow up with other PRs.

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -84,6 +84,18 @@ type RunPolicy struct {
 	// SchedulingPolicy defines the policy related to scheduling, e.g. gang-scheduling
 	// +optional
 	SchedulingPolicy *SchedulingPolicy `json:"schedulingPolicy,omitempty"`
+
+	// suspend specifies whether the MPIJob controller should create Pods or not.
+	// If a MPIJob is created with suspend set to true, no Pods are created by
+	// the MPIJob controller. If a MPIJob is suspended after creation (i.e. the
+	// flag goes from false to true), the MPIJob controller will delete all
+	// active Pods associated with this MPIJob. Also, it will suspend the
```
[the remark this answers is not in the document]
mimowo: Done

pkg/apis/kubeflow/v2beta1/types.go
mimowo: Yes, I was thinking about this. Still I would suggest in a follow up PR to separate the new logic and the refactoring.

tenzen-y: Sounds good.

tenzen-y:
@mimowo Thanks for the great work!
/lgtm

/assign @alculquicondor

alculquicondor (commented):
Can you add a unit test?

Also, I'm not convinced of the value of an E2E test over unit+integration. Do you have a particular justification?

pkg/controller/mpi_job_controller.go
```diff
@@ -905,6 +936,19 @@ func (c *MPIJobController) updateMPIJobStatus(mpiJob *kubeflow.MPIJob, launcher
 	if err != nil {
 		return fmt.Errorf("checking launcher pods running: %w", err)
 	}
+	if isMPIJobSuspended(mpiJob) {
+		// it is suspended now
+		if updateMPIJobConditions(mpiJob, kubeflow.JobSuspended, v1.ConditionTrue, "MPIJobSuspended", "MPIJob suspended") {
```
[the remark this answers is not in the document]
mimowo: Yeah, I thought so to, but this seems to be a convention here. For now, I stick to the convention but added the reason to the list.

pkg/controller/mpi_job_controller.go
```diff
@@ -1304,7 +1351,7 @@ func (c *MPIJobController) newWorker(mpiJob *kubeflow.MPIJob, index int) *corev1
 }
 
 func (c *MPIJobController) newLauncherJob(mpiJob *kubeflow.MPIJob) *batchv1.Job {
-	return &batchv1.Job{
+	job := &batchv1.Job{
```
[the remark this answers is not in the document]
mimowo: Good point. IIUC, this affects only Kueue configurations when ManageJobsWithoutQueueName=true, which is non-default. We could have an annotation, yes, or just do not manage by Kueue any Job objects which have OwnerReference to another object managed by Kueue.

mimowo:
> Can you add a unit test?

Done, ended up adding 3 actually: for creating suspended MPIJob, suspending if running and resuming.
One thing is that to write the unit test for resuming I had to refactor the code a little bit to inject a fake clock in tests.
Also, the test for suspending a running MPIJob revealed that I was requiring two syncs - one to clean up the pod workers
and one to update the MPIJob status. Now, I do these steps in one sync.

> Also, I'm not convinced of the value of an E2E test over unit+integration. Do you have a particular justification?

I think of two reasons:
1. the test abstracts out the implementation details, thus is better at documenting what the feature is about. For example, it abstracts out when the service and other auxiliary objects are created. Thus, having such a test allows us to do refactoring and have confidence that the feature works before the unit, or integration tests are adjusted.
2. it makes us more confident about race conditions. For example, suspending or resuming an MPIJob triggers
suspending or resuming the Launcher Job which happens asynchronously. We don't really test the interaction
between the launcher job and the MPIJob at other layers of testing.

manifests/base/crd.yaml
```diff
@@ -94,7 +96,7 @@ spec:
                   properties:
                     type:
                       type: string
-                      enum: ["Created", "Running", "Restarting", "Succeeded", "Failed"]
+                      enum: ["Created", "Running", "Restarting", "Succeeded", "Suspended", "Failed"]
```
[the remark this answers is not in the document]
alculquicondor: Uhm... then make sure that `make generate` doesn't override it.

pkg/controller/mpi_job_controller.go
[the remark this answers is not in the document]
alculquicondor: > or just do not manage by Kueue any Job objects which have OwnerReference to another object managed by Kueue.

The dependency could be indirect. And we don't want to spend a GET call to obtain such information.

manifests/base/crd.yaml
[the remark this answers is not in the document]
mimowo: it doesn't, checked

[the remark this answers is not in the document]
alculquicondor: @tenzen-y is this expected?
Or is this related to your other PR?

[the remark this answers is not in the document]
alculquicondor: We can leave this to #510

pkg/controller/mpi_job_controller_test.go
```diff
@@ -751,6 +765,209 @@ func TestShutdownWorker(t *testing.T) {
...
+			mpiJobCopy.Status.ReplicaStatuses = map[common.ReplicaType]*common.ReplicaStatus{
+				common.ReplicaType(kubeflow.MPIReplicaTypeLauncher): {},
+				common.ReplicaType(kubeflow.MPIReplicaTypeWorker):   {},
+			}
+			msg := fmt.Sprintf("MPIJob %s/%s is created.", mpiJob.Namespace, mpiJob.Name)
+			updateMPIJobConditions(mpiJobCopy, common.JobCreated, v1.ConditionTrue, mpiJobCreatedReason, msg)
+			updateMPIJobConditions(mpiJobCopy, kubeflow.JobSuspended, v1.ConditionTrue, mpiJobSuspendedReason, "MPIJob suspended")
+			msg = fmt.Sprintf("MPIJob %s/%s is suspended.", mpiJob.Namespace, mpiJob.Name)
+			updateMPIJobConditions(mpiJobCopy, common.JobRunning, v1.ConditionFalse, mpiJobSuspendedReason, msg)
+			f.expectUpdateMPIJobStatusAction(mpiJobCopy)
```
alculquicondor, line 813: expect zero workers

test/e2e/mpi_job_test.go
```diff
@@ -211,12 +231,31 @@ var _ = ginkgo.Describe("MPIJob", func() {
 	})
 })
 
+func resumeJob(mpiJob *kubeflow.MPIJob) *kubeflow.MPIJob {
```
alculquicondor, line 234: For follow up: accept contexts in all these functions

manifests/base/crd.yaml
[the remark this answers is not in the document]
tenzen-y: Yes, this is a known issue, and this will be fixed by #510.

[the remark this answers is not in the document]
mimowo: Yeah, but it is needed for the e2e test (which I believe is worth adding).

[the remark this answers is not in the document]
alculquicondor: Yeah, keep the manual change, and #510 should automatize it.

pkg/controller/mpi_job_controller_test.go
mimowo: The code above checks that already:
```golang
			mpiJobCopy.Status.ReplicaStatuses = map[common.ReplicaType]*common.ReplicaStatus{
				common.ReplicaType(kubeflow.MPIReplicaTypeLauncher): {},
				common.ReplicaType(kubeflow.MPIReplicaTypeWorker):   {},
			}
```
It does not specify `Active`, meaning it checks the value is 0.

The workers were never created in this scenario, so I cannot assert on delete actions.

test/e2e/mpi_job_test.go
mimowo: Will open a PR directly after this, doesn't seem it requires Issue for later.

mimowo: Similarly, going to open a PR to copy the MPIJob conditions from common.

tenzen-y: @mimowo I opened the PR to copy the MPIJob conditions to this repo in #514 since I faced issues caused by the MPIJob conditions in #510.

pkg/controller/mpi_job_controller_test.go
alculquicondor: but you can assert that there were no pods were created

the status is not the same as pods being created, necessarily

alculquicondor: oh I see, so it's implicitly checked

terrytangyuan (approved):
Thanks!

/lgtm
/approve

terrytangyuan:
The other PR got merged first so this one will need to resolve conflicts :-)

alculquicondor:
Still lgtm

Issue #518, opened 2023-02-05
Respect SchedulingPolicy


tenzen-y:
/kind feature

Currently, the mpi-operator does not support the `Runpolicy.SchedulingPolicy`. For example, we don't respect the `SchedulingPolicy.MinAvailable` when we create the PodGroup in the following:

https://github.com/kubeflow/mpi-operator/blob/5f1914bfb29b9b94db209e620a3b2bf7e5ca7d9f/pkg/controller/mpi_job_controller.go#L1291-L1316

Also, supporting the SchedulingPolicy, we can set the various PodGroup parameters for the coscheduling-plugin.

```go
// PodGroupSpec represents the template of a pod group.
type PodGroupSpec struct {
	// MinMember defines the minimal number of members/tasks to run the pod group;
	// if there's not enough resources to start all tasks, the scheduler
	// will not start anyone.
	MinMember int32 `json:"minMember,omitempty"`

	// MinResources defines the minimal resource of members/tasks to run the pod group;
	// if there's not enough resources to start all tasks, the scheduler
	// will not start anyone.
	MinResources v1.ResourceList `json:"minResources,omitempty"`

	// ScheduleTimeoutSeconds defines the maximal time of members/tasks to wait before run the pod group;
	ScheduleTimeoutSeconds *int32 `json:"scheduleTimeoutSeconds,omitempty"`
}
```

https://github.com/kubernetes-sigs/scheduler-plugins/blob/f996e5caf6c77d521d574186dca793e351c45413/apis/scheduling/v1alpha1/types.go#L139-L153

tenzen-y:
/assign
Blocking #500.

PR #520, opened 2023-02-08
Respect SchedulingPolicy

tenzen-y:
Signed-off-by: Yuki Iwai <yuki.iwai.tz@gmail.com>

I changed the logic for the gang-scheduling so that the mpi-operator respects SchedulingPolicy when creating PodGroup.

Fixes: #518 

/assign @alculquicondor


alculquicondor:
Can you summarize the changes for somebody that isn't familiar with volcano? :sweat_smile:

tenzen-y:
> Can you summarize the changes for somebody that isn't familiar with volcano? 😅

Sure. I will send a ping to you once this PR description is ready.

pkg/controller/podgroup.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if l := replicas[kubeflow.MPIReplicaTypeLauncher]; l != nil && len(l.Template.Spec.PriorityClassName) != 0 {
+		return l.Template.Spec.PriorityClassName
+	} else if w := replicas[kubeflow.MPIReplicaTypeWorker]; w != nil && len(w.Template.Spec.PriorityClassName) != 0 {
+		return w.Template.Spec.PriorityClassName
+	}
+	return ""
+}
+
+func (c *MPIJobController) calcPGMinResources(
```
alculquicondor, line 89: add a comment for what's happening here

pkg/controller/mpi_job_controller.go
```diff
@@ -388,13 +401,20 @@ func (c *MPIJobController) Run(threadiness int, stopCh <-chan struct{}) error {
...
+		c.jobSynced,
+		c.podSynced,
+		c.priorityClassSynced,
+		c.mpiJobSynced,
 	}
 	if c.gangSchedulerName != "" {
-		if ok := cache.WaitForCacheSync(stopCh, c.podgroupsSynced); !ok {
-			return fmt.Errorf("failed to wait for podgroup caches to sync")
-		}
+		synced = append(synced, c.podgroupsSynced)
```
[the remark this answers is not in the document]
tenzen-y: Yes, that's right. 
It would be good to sync priorityclass only when the gangSchedulerName is set, same as priorityclass.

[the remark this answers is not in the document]
tenzen-y: I will change the above logic.

pkg/controller/podgroup.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		if len(schedulingPolicy.Queue) != 0 {
+			queueName = schedulingPolicy.Queue
+		}
+		if schedulingPolicy.MinResources != nil {
+			minResources = schedulingPolicy.MinResources
+		}
+	}
+	if minResources == nil {
+		minResources = c.calcPGMinResources(minMember, mpiJob.Spec.MPIReplicaSpecs)
```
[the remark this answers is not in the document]
tenzen-y: Yes. Previously, we could not use `minResources` since we don't calculate and set that:

https://github.com/kubeflow/mpi-operator/blob/c21942d1e27d196d06078f67f39a6f7a0e990dee/pkg/controller/mpi_job_controller.go#L1299-L1324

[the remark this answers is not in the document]
tenzen-y: If we configurable the `minResources`, scheduling part of Workers to Nodes is useful when the number of Workers is huge.

pkg/controller/podgroup.go
```diff
@@ -0,0 +1,168 @@
...
+// newPodGroup creates a new PodGroup for an MPIJob
+// resource. It also sets the appropriate OwnerReferences on the resource so
+// handleObject can discover the MPIJob resource that 'owns' it.
+func (c *MPIJobController) newPodGroup(mpiJob *kubeflow.MPIJob) *volcanov1beta1.PodGroup {
+	minMember := workerReplicas(mpiJob) + 1
+	queueName := mpiJob.Annotations[volcanov1beta1.QueueNameAnnotationKey]
+	var minResources *corev1.ResourceList
+	if schedulingPolicy := mpiJob.Spec.RunPolicy.SchedulingPolicy; schedulingPolicy != nil {
+		if schedulingPolicy.MinAvailable != nil {
+			minMember = *schedulingPolicy.MinAvailable
```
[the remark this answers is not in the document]
tenzen-y: > what if volcano schedules a number of pods that is less than workerReplicas(mpiJob)?
The mpirun command might get stuck waiting for all the workers (unless it's elastic horovod that has native support for changing number of workers).

In that case, volcano schedules all Pods. If the cluster does not have enough resources for all Pods, some pods will be marked as `Pending`.

> We need a way to restrict the usage of this field or have clear documentation that this is not supported for all kinds of MPI applications.

I agree. I think we have 2 options. Since we don't have fields if the MPIJob uses elastic horovod. 

1. As you say, have clear documentation.
2. Add a new member, 'Elastic bool `json:elastic,omitempty`' into `Runpolicy` to represent if MPIJob uses elastic mode. And then, we add validations for `minMember` and `minResources`.

I like the second option. What do you think? @alculquicondor

tenzen-y:
@alculquicondor Updated PR description.

pkg/controller/podgroup.go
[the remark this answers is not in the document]
alculquicondor: Are there similar fields in the training operator?

pkg/controller/mpi_job_controller.go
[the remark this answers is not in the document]
alculquicondor: still not changed

pkg/controller/podgroup.go
```diff
@@ -0,0 +1,168 @@
...
+
+type replicaPriority struct {
+	priority int32
+
+	common.ReplicaSpec
+}
+
+// newPodGroup creates a new PodGroup for an MPIJob
+// resource. It also sets the appropriate OwnerReferences on the resource so
+// handleObject can discover the MPIJob resource that 'owns' it.
```
alculquicondor, line 38: explain how minResources is calculated.

What does volcano use that field? If it's all added up, how does it know whether they are consumed by one or multiple pods?

pkg/controller/podgroup.go
alculquicondor: Still waiting.

What does volcano use that field? If it's all added up, how does it know whether they are consumed by one or multiple pods?

pkg/controller/podgroup.go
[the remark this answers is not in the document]
tenzen-y: No, most customjob-controllers have no fields on whether CustomJob uses elastic mode. This means users can set any value to `schedulingPolicy.MinMember`. The pytorchjob-controller has `ElasticPolicy`, although that field doesn't affect schedulingPolicy.

https://github.com/kubeflow/training-operator/blob/6ce98387a1a68b79365495a6344f7f19b48557d3/pkg/apis/kubeflow.org/v1/pytorch_types.go#L79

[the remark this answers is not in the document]
alculquicondor: in that case, I think we should proceed with option 1 for now.

pkg/controller/podgroup.go
tenzen-y: > explain how minResources is calculated.

Sure.

> What does volcano use that field? If it's all added up, how does it know whether they are consumed by one or multiple pods?

I will answer in another thread.

pkg/controller/podgroup.go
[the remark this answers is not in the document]
tenzen-y: > in that case, I think we should proceed with option 1 for now.

I agree. Maybe, we can add documentation to https://www.kubeflow.org/docs/components/training/job-scheduling/.

> That said, there is currently no documentation about SchedulingPolicy overall 

You're right...

pkg/controller/podgroup.go
tenzen-y: > What does volcano use that field? If it's all added up, how does it know whether they are consumed by one or multiple pods?

I tried to dive into the volcano. The volcano has the Queue resource like ClusterQueue in Kueue. 
If Queue doesn't have enough resources requested by `minResources`  in PodGroup, volcano-scheduler doesn't schedule all Pods to Nodes.

Does that make sense?

alculquicondor: makes sense.... it looks like a quota system just like kueue. I wonder if they would be interested in reviewing this PR. Do you know anybody?

tenzen-y: I remember @zw0610 is familiar with volcano. And @zw0610 is a reviewer in this repository :)

tenzen-y: IMO, as a result of investigating the volcano, maybe we don't need the `calcPGMinResources` function since `minResources` don't affect gang-scheduling. And we may be able to propagate `schedulingPolicy.minResources` to PodGroup.

tenzen-y: @zw0610 Let us know what you think.

tenzen-y: Maybe, @shinytang6 is also interested in this PR.
cc: @shinytang6

tenzen-y: @alculquicondor They don't seem to have enough bandwidth to review this PR.
So, I'm thinking of removing `func calcPGMinResources` and then passing `SchedulingPolicy.MinResources` to `PodGroup.Spec.MinResources`.

Also, it would be good to keep tracking this.

alculquicondor: SGTM, we don't want to introduce functionality we aren't sure about.

tenzen-y: > we don't want to introduce functionality we aren't sure about.

I agree.

pkg/controller/podgroup.go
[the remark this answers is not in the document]
tenzen-y: > So we need to clarify it in the documentation anyways.

I agree.

Probably, we can discuss if we add an `elastic` field in other places (issue or pr).
In this PR, we can just pass `SchedulingPolicy.MinAvailable` to `PodGroup.Spec.MinMember` and then add docs about that.

pkg/controller/mpi_job_controller.go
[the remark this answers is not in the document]
tenzen-y: Done.

pkg/controller/podgroup.go
tenzen-y: Done.

tenzen-y:
@alculquicondor I have addressed your comments. Please take another look.

cmd/mpi-operator/app/server.go
```diff
@@ -46,6 +46,7 @@ import (
 
 	"github.com/kubeflow/mpi-operator/cmd/mpi-operator/app/options"
 	mpijobclientset "github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned"
+	kubeflowScheme "github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned/scheme"
```
alculquicondor, line 49:
```suggestion
	kubeflowscheme "github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned/scheme"
```

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -54,11 +54,35 @@ const (
...
 // job, for example `minAvailable` for gang-scheduling.
 type SchedulingPolicy struct {
-	MinAvailable           *int32           `json:"minAvailable,omitempty"`
-	Queue                  string           `json:"queue,omitempty"`
-	MinResources           *v1.ResourceList `json:"minResources,omitempty"`
-	PriorityClass          string           `json:"priorityClass,omitempty"`
-	ScheduleTimeoutSeconds *int32           `json:"scheduleTimeoutSeconds,omitempty"`
+	// MinAvailable defines the minimal number of member to run the PodGroup.
+	// If the gang-scheduling is set to the volcano,
+	// input is passed to `.spec.mimMember` in PodGroup for the volcano.
```
alculquicondor, line 59: Explain that this will only work if the MPI application supports resizing.

pkg/controller/mpi_job_controller.go
```diff
@@ -364,6 +376,13 @@ func NewMPIJobControllerWithClock(
 		UpdateFunc: controller.handleObjectUpdate,
 		DeleteFunc: controller.handleObject,
 	})
+	if priorityClassInformer != nil {
+		priorityClassInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    controller.handleObject,
+			UpdateFunc: controller.handleObjectUpdate,
+			DeleteFunc: controller.handleObject,
```
alculquicondor, line 383: Does it make sense to add an event handler for priorityClass? They are not owned by the MPIJob.

pkg/controller/podgroup.go
alculquicondor: add in the comment the fields that this function populates and how

pkg/controller/podgroup.go
```diff
@@ -0,0 +1,88 @@
...
+//  2. .spec.mpiReplicaSecs[Launcher].template.spec.priorityClassName
+//  3. .spec.mpiReplicaSecs[Worker].template.spec.priorityClassName
+//
+// However, if the calculated priorityClass name is missing in the cluster, return an empty.
+func (c *MPIJobController) calcPriorityClassName(
+	replicas map[kubeflow.MPIReplicaType]*common.ReplicaSpec,
+	schedulingPolicy *kubeflow.SchedulingPolicy,
+) string {
+	if schedulingPolicy != nil && len(schedulingPolicy.PriorityClass) != 0 &&
+		c.priorityClassExist(schedulingPolicy.PriorityClass) {
```
alculquicondor, line 72: Should MPIJob care if it exists? I think it's up to Volcano to decide what to do.

pkg/apis/kubeflow/v2beta1/types.go
tenzen-y: Sounds good.

pkg/controller/mpi_job_controller.go
tenzen-y: You're right.

pkg/controller/podgroup.go
tenzen-y: Makes sense.

cmd/mpi-operator/app/server.go
tenzen-y: Done.

pkg/apis/kubeflow/v2beta1/types.go
tenzen-y: Done.

pkg/controller/podgroup.go
tenzen-y: We can leave it to the volcano.

alculquicondor: Then let's cleanup the informer

tenzen-y: Done.

pkg/controller/mpi_job_controller.go
tenzen-y: Done.

pkg/controller/podgroup.go
tenzen-y: Done.

tenzen-y:
@alculquicondor Updated. PTAL.

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -54,11 +54,36 @@ const (
...
+	// Also, this parameter will function properly only when we use elastic training (e.g., Elastic Horovod).
```
alculquicondor, line 60:
```suggestion
	// When using this field, you need to make sure the application supports resizing (e.g., Elastic Horovod).
```

pkg/apis/kubeflow/v2beta1/types.go
alculquicondor: If not set, it defaults to the number of workers

tenzen-y:
@alculquicondor I addressed your suggestions and squashed commits into one.

tenzen-y:
Also, I will add docs for the schedulingPolicy to https://www.kubeflow.org/.

tenzen-y:
@alculquicondor Created https://github.com/kubeflow/website/pull/3453.

alculquicondor:
oops, no, I just forgot to approve
/approve

Issue #505, opened 2023-01-24
Adapt mpi-operator v2 to Kueue


tenzen-y:
Many users hope mpi-operator v2 adapts [Kueue](https://github.com/kubernetes-sigs/kueue).

Blocked by #504 https://github.com/kubernetes-sigs/kueue/issues/369
Kueue side issue: https://github.com/kubernetes-sigs/kueue/issues/65

/kind feature

alculquicondor:
I suppose you mean creating and syncing a Workload object for each MPIJob.

Do you think we should have this code in mpi-operator or in kueue?

tenzen-y:
> I suppose you mean creating and syncing a Workload object for each MPIJob.

Yes, I meant we implement a controller for Workload resources like the job-controller.
https://github.com/kubernetes-sigs/kueue/blob/10d35322c252c2724467cf4617e79e94e1bd0c8a/pkg/controller/workload/job/job_controller.go

> Do you think we should have this code in mpi-operator or in kueue?

IIUC, kueue is designed not to hold third-party dependencies.
So we might need to add that code into mpi-operator, right?

alculquicondor:
Yes, we could have this controller in any of the repos. Wherever it goes, we should have them for kueue+mpi-operator. Both repos have the setup for E2E tests.

Having it in kueue might be better for the time being as things are still changing. But having it in mpi-operator serves as proof that custom jobs don't have to be in tree (and we can add references from the kueue README).

I personally prefer to have it in mpi-operator if the OWNERS don't mind (cc @terrytangyuan)

@ahg-g, wdyt?

terrytangyuan:
I am okay with either approach.

ahg-g:
If we put it in mpi-operator repo, I supposed it will run as a reconciler within the same binary, not yet another operator, correct?

alculquicondor:
Yes, same binary, possibly guarded by a command line flag to enable

tenzen-y:
> Having it in kueue might be better for the time being as things are still changing. But having it in mpi-operator serves as proof that custom jobs don't have to be in tree (and we can add references from the kueue README).

I'm also fine either way. 

If we select the latter, my concern is when kueue will stop serving that controller and donate that controller to the mpi-operator repo. 

If kueue keeps serving that controller, we might face why kueue does not provide other controllers (e.g., Ray, Argo, Spark, and more).

However, if we select the former, the mpi-operator may face API changes of kueue since kueue has alpha status as you say.

[...]

PR #521, opened 2023-02-09
WIP: Use the Kueue ignore annotation

mimowo:
WIP because we want the Kueue PR (https://github.com/kubernetes-sigs/kueue/pull/558) to merge first to make sure the annotation name is correct.

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -118,6 +118,9 @@ const (
 
 	openMPISlotsEnv  = "OMPI_MCA_orte_set_default_slots"
 	intelMPISlotsEnv = "I_MPI_PERHOST"
+
+	// Annotation instructing Kueue to ignore a workload for scheduling.
+	KueueIgnoreAnnotation = "kueue.x-k8s.io/ignore"
 )
 
 var (
@@ -1386,6 +1389,9 @@ func (c *MPIJobController) newLauncherJob(mpiJob *kubeflow.MPIJob) *batchv1.Job 
 			Namespace: mpiJob.Namespace,
 			Labels: map[string]string{
 				"app": mpiJob.Name,
+			},
+			Annotations: map[string]string{
+				KueueIgnoreAnnotation: "true",
 			},
 			OwnerReferences: []metav1.OwnerReference{
 				*metav1.NewControllerRef(mpiJob, kubeflow.SchemeGroupVersionKind),
--- a/pkg/controller/mpi_job_controller_test.go
+++ b/pkg/controller/mpi_job_controller_test.go
@@ -1206,6 +1206,9 @@ func TestNewLauncherAndWorker(t *testing.T) {
 					Labels: map[string]string{
 						"app": "foo",
 					},
+					Annotations: map[string]string{
+						KueueIgnoreAnnotation: "true",
+					},
 				},
 				Spec: batchv1.JobSpec{
 					Template: corev1.PodTemplateSpec{
@@ -1368,6 +1371,9 @@ func TestNewLauncherAndWorker(t *testing.T) {
 					Namespace: "foo",
 					Labels: map[string]string{
 						"app": "bar",
+					},
+					Annotations: map[string]string{
+						KueueIgnoreAnnotation: "true",
 					},
 				},
 				Spec: batchv1.JobSpec{


pkg/controller/mpi_job_controller.go
```diff
@@ -118,6 +118,9 @@ const (
 
 	openMPISlotsEnv  = "OMPI_MCA_orte_set_default_slots"
 	intelMPISlotsEnv = "I_MPI_PERHOST"
+
+	// Annotation instructing Kueue to ignore a workload for scheduling.
+	KueueIgnoreAnnotation = "kueue.x-k8s.io/ignore"
```
tenzen-y, line 123: Can we merge this change into the master branch after the release of kueue v0.3.0, including https://github.com/kubernetes-sigs/kueue/pull/558? 

Also, it would be good to import this annotation from kueue instead of hard coding.


cc: @alculquicondor

mimowo: > Can we merge this change into the master branch after the release of kueue v0.3.0, including [kubernetes-sigs/kueue#558](https://github.com/kubernetes-sigs/kueue/pull/558)?

We could wait, but I don't see a clear benefit. Once the Kueue PR is merged the rename of the annotation will be unlikely. Also, adding the annotation in case running with Kueue 0.2.0 will be harmless and the Kueue will simply ignore it. One advantage of merging early is that we will be able to continue the remaining Kueue work (the operator to suspend / unsuspend the MPIJob) using master of the mpi-operator, and not a branch.

> Also, it would be good to import this annotation from kueue instead of hard coding.

I agree, however I don't think we want MPI operator to depend on the entire Kueue, so we would need to first extract a small kueue-api library with the API for integrations. I guess it could be providing the interfaces for integration similar to those suggested in the KEP: https://github.com/kubernetes-sigs/kueue/pull/537. However, I think it would be good to first do one or two example integrations to see how the integration looks like and then extract the library, but possibly for 0.4.0.

tenzen-y: > One advantage of merging early is that we will be able to continue the remaining Kueue work (the operator to suspend / unsuspend the MPIJob) using master of the mpi-operator, and not a branch.

IIRC, we are planning to add workload-controller for MPIJob (like https://github.com/kubernetes-sigs/kueue/tree/main/pkg/controller/workload/job) to mpi-operator repository. So it seems that we will be able to continue the remaining Kueue work without merging this PR. WDYT?

ref: https://github.com/kubeflow/mpi-operator/issues/505

tenzen-y: > I guess it could be providing the interfaces for integration similar to those suggested in the KEP: https://github.com/kubernetes-sigs/kueue/pull/537. However, I think it would be good to first do one or two example integrations to see how the integration looks like and then extract the library, but possibly for 0.4.0.

It sounds good to me.

mimowo: > IIRC, we are planning to add workload-controller for MPIJob (like https://github.com/kubernetes-sigs/kueue/tree/main/pkg/controller/workload/job) to mpi-operator repository. So it seems that we will be able to continue the remaining Kueue work without merging this PR. WDYT?



> IIRC, we are planning to add workload-controller for MPIJob (like https://github.com/kubernetes-sigs/kueue/tree/main/pkg/controller/workload/job) to mpi-operator repository. So it seems that we will be able to continue the remaining Kueue work without merging this PR. WDYT?
> 
> ref: #505

Technically it could be either in mpi-operator or in Kueue, I don't have a strong opinion where. The argument for putting it into Kueue is that then we have a clear view of how the integrations look like and so we may have a better idea of how the common library should look like. 

Still, I don't see a clear benefit in holding the annotation task. It should also not be influencing the decision about the repo. 

cc: @mwielgus.

tenzen-y: > Technically it could be either in mpi-operator or in Kueue, I don't have a strong opinion where. The argument for putting it into Kueue is that then we have a clear view of how the integrations look like and so we may have a better idea of how the common library should look like.

I see. It looks like you have a point. Let me know what others think. cc: @alculquicondor @ahg-g @kerthcet

However, If we add a workload-controller for MPIJob to the kueue repository, it might be better to implement the logic to propagate `.metadata.annotations` of MPIJob to Launcher and Worker instead of hard coding `kueue.x-k8s.io/ignore` (this means the mpi-operator propagates annotations of MPIJob to Launcher batch/v1 Job and Worker Pods). And then, a workload-controller injects `kueue.x-k8s.io/ignore` to annotations of MPIJob.

Since it would be unnecessary for users who don't use kueue.

WDYT?

mimowo: > However, If we add a workload-controller for MPIJob to the kueue repository, it might be better to implement the logic to propagate `.metadata.annotations` of MPIJob to Launcher and Worker instead of hard coding `kueue.x-k8s.io/ignore` (this means the mpi-operator propagates annotations of MPIJob to Launcher batch/v1 Job and Worker Pods). And then, a workload-controller injects `kueue.x-k8s.io/ignore` to annotations of MPIJob.

Let me check if I understand correctly. You suggest that the mpi-operator (in this repo) should propagate all annotations it got set by the workload-controller for MPIJob in Kueue down to the Launcher. However, it would mean that we need to add the `kueue.x-k8s.io/ignore` to MPIJob, and this is not what we want. We want the MPIJob to be scheduled by Kueue, not ignored. Thus we need to have the annotation only at the Launcher level, but not at the MPIJob level.

tenzen-y: > Let me check if I understand correctly. You suggest that the mpi-operator (in this repo) should propagate all annotations it got set by the workload-controller for MPIJob in Kueue down to the Launcher. 

Yes, that's right.  

> However, it would mean that we need to add the kueue.x-k8s.io/ignore to MPIJob, and this is not what we want. We want the MPIJob to be scheduled by Kueue, not ignored. Thus we need to have the annotation only at the Launcher level, but not at the MPIJob level.

As you say, it might be weird that Workers have `kueue.x-k8s.io/ignore` annotations. 
So if the kueue repo has a workload-controller for MPIJob, this PR change might be a better way.

Thanks for clarifying.

mimowo: IIUC It means no changes required in the mpi-operator, so closing the PR.

tenzen-y: Sounds good. I prefer the way to insert annotation by webhook on kueue side.

alculquicondor: Yes, being able to integrate without having to inject kueue-specific logic is a great benefit!

PR #530, opened 2023-02-22
Stop using e2e tag

tenzen-y:
If we add build tags to e2e test codes, `go fmt` and `go vet` ignores those files. So I removed the e2e tag so that we could format those codes.

Ref: https://github.com/golang/go/issues/4007


alculquicondor (commented):
/lgtm
/approve

PR #533, opened 2023-02-28
Stop using the kustomize vars feature

tenzen-y:
Since kustomize v5, the `vars` feature has been deprecated. So I replaced the `vars` with the kustomize patch.

> The vars field was deprecated in v5.0.0. This field will never be removed from the kustomize.config.k8s.io/v1beta1 Kustomization API, but it will not be included in the kustomize.config.k8s.io/v1 Kustomization API. When Kustomization v1 is available, we will announce the deprecation of the v1beta1 version. There will be at least two releases between deprecation and removal of Kustomization v1beta1 support from the kustomize CLI, and removal itself will happen in a future major version bump.

https://github.com/kubernetes-sigs/cli-experimental/blob/7b72bd123ddad23c20bfc6607a039a08717b6934/site/content/en/references/kustomize/kustomization/vars/_index.md

### DIFF
--- a/deploy/v2beta1/mpi-operator.yaml
+++ b/deploy/v2beta1/mpi-operator.yaml
@@ -8269,19 +8269,6 @@ subjects:
   name: mpi-operator
   namespace: mpi-operator
 ---
-apiVersion: v1
-data:
-  lock-namespace: mpi-operator
-kind: ConfigMap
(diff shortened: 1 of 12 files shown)


alculquicondor (commented):
/lgtm
/approve

Issue #500, opened 2023-01-12
Support coscheduling plugin


tenzen-y:
/kind feature

MPI Operator now supports the all-or-nothing semantic, queuing logic, and more features for batch workload by the Volcano.
Although, I think the maintenance cost for Volcano is a bit high for users who want to use only all-or-nothing semantic.

So I would like to support that semantic by [coscheduling plugin](https://github.com/kubernetes-sigs/scheduler-plugins/blob/610e2314adeac51ed75699757c55238fe26d99ec/pkg/coscheduling/README.md).

Supporting the coscheduling plugin, users could use that semantic without additional components.

@alculquicondor @gaocegege @terrytangyuan @zw0610 WDYT?

---

Tasks:

- [x] Implement the logic to support the coscheduling plugin
- [x] Add E2E for gang-scheduling using the scheduler plugins

zw0610:
I support this feature enhancement. Meanwhile, will mpi-operator still keep support for Volcano?

tenzen-y:
> I support this feature enhancement. Meanwhile, will mpi-operator keep support for Volcano?

Yes, I was thinking of adding support for the coscheduling plugin, not replacing volcano.

zw0610:
I was trying to support this in the following pull requests:
https://github.com/kubeflow/common/pull/185
https://github.com/kubeflow/training-operator/pull/1526

But there are still some version conflict between the coscheduling plugin and training operator when the last time I try to push forward this feature.

I will close these two when new pull requests are submitted.

tenzen-y:
@zw0610 I created an issue to support the coscheduling plugin for training-operator at https://github.com/kubeflow/training-operator/issues/1722 since I missed those PRs.

As I can see mpi-operator v2, I think we can support the scheduler plugin by simply implementing it in this repository since mpi-operator v2 doesn't depend on kubeflow/common.

alculquicondor:
Note that the coscheduling plugin is still not part of a standard kubernetes installation. The enhancement is tracked here, but it's currently stalled https://github.com/kubernetes/enhancements/issues/3370

I'm not sure if the design would change much. Perhaps @denkensk and @Huang-Wei could provide some updates.

> But there are still some version conflict between the coscheduling plugin and training operator when the last time I try to push forward this feature.

Can you expand on this?

tenzen-y:
[...]

PR #538, opened 2023-03-20
Support the coscheduling plugin of scheduler-plugins

tenzen-y:
I added the implementation for the coscheduling plugin of kubernetes-sigs/scheduler-plugins.

ref: https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/coscheduling

So I modified the following:

- Added PodGroupControl interface inspired by https://github.com/kubeflow/common/blob/4dbf18c85417b54158914d1d01c1744b4e3542f2/pkg/controller.v1/control/pod_control.go#L53-L65.
- Added VolcanoCtrl implemented PodGroupCtrl interface. The logic is the same as before.
- Added SchedulerPluginsCtrl implemented PodGroupCtrl interface.

Follow ups:
- Add E2E tests

Part-of: #500


tenzen-y:
I will add the integration tests to this PR.

/assign @alculquicondor

cmd/mpi-operator/app/server.go
```diff
@@ -136,37 +137,43 @@ func Run(opt *options.ServerOption) error {
...
-		if opt.GangSchedulingName != "" {
-			podgroupsInformer = volcanoInformerFactory.Scheduling().V1beta1().PodGroups()
+		var podGroupCtrl controllersv1.PodGroupControl
+		if opt.GangSchedulingName == options.GangSchedulerVolcano {
+			podGroupCtrl = controllersv1.NewVolcanoCtrl(volcanoClientSet, volcanoInformerFactory)
+		} else if len(opt.GangSchedulingName) != 0 {
+			// Use scheduler-plugins as a default gang-scheduler.
+			podGroupCtrl = controllersv1.NewSchedulerPluginsCtrl(schedClientSet, schedInformerFactory, opt.GangSchedulingName)
 		}
+
```
tenzen-y, line 160: Use scheduler-plugins as a default gang-scheduler.

go.mod
```diff
@@ -9,15 +9,16 @@ require (
...
+	k8s.io/api v0.25.7
+	k8s.io/apimachinery v0.25.7
+	k8s.io/apiserver v0.25.7
+	k8s.io/client-go v0.25.7
+	k8s.io/code-generator v0.25.7
 	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20230109183929-3758b55a6596
 	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
 	sigs.k8s.io/controller-runtime v0.13.1
+	sigs.k8s.io/scheduler-plugins v0.25.7
```
tenzen-y, line 21: We use the newest version since this version introduces a new API group, `scheduling.x-k8s.io`.

alculquicondor: there is no 0.26?

pkg/controller/mpi_job_controller.go
```diff
@@ -28,6 +28,13 @@ import (
 	"strconv"
 	"time"
 
+	common "github.com/kubeflow/common/pkg/apis/common/v1"
```
alculquicondor, line 31: just one import group now?

pkg/controller/mpi_job_controller_test.go
```diff
@@ -233,16 +260,16 @@ func (f *fixture) newController(gangSchedulerName string, clock clock.WithTicker
 	return c, i, k8sI
 }
 
-func (f *fixture) run(mpiJobName string) {
-	f.runWithClock(mpiJobName, clock.RealClock{})
+func (f *fixture) run(mpiJobName, gangSchedulerName string) {
```
alculquicondor, line 263: should the scheduler name be part of the fixture?

pkg/controller/podgroup.go
```diff
@@ -58,19 +119,224 @@ func newPodGroup(mpiJob *kubeflow.MPIJob) *volcanov1beta1.PodGroup {
...
+		order[1].Replicas = pointer.Int32(*minMember - 1)
+	}
+
+	minResources := corev1.ResourceList{}
+	for _, rp := range order {
+		if rp.Replicas == nil {
+			continue
+		}
+		for i := int32(0); i < *rp.Replicas; i++ {
+			for _, c := range rp.Template.Spec.Containers {
```
alculquicondor, line 316: what about initContainers?

cmd/mpi-operator/app/options/options.go
```diff
@@ -62,7 +67,9 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 	fs.IntVar(&s.MonitoringPort, "monitoring-port", 0,
 		`Endpoint port for displaying monitoring metrics. It can be set to "0" to disable the metrics serving.`)
 
-	fs.StringVar(&s.GangSchedulingName, "gang-scheduling", "", "Set gang scheduler name if enable gang scheduling.")
+	fs.StringVar(&s.GangSchedulingName, "gang-scheduling", "",
+		`Set gang scheduler name if enable gang scheduling.
+                Now Supporting default-scheduler, volcano, and scheduler-plugins.`)
```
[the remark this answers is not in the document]
tenzen-y: That makes sense.

cmd/mpi-operator/app/server.go
```diff
@@ -136,37 +138,48 @@ func Run(opt *options.ServerOption) error {
 		var kubeInformerFactory kubeinformers.SharedInformerFactory
 		var kubeflowInformerFactory informers.SharedInformerFactory
 		var volcanoInformerFactory volcanoinformers.SharedInformerFactory
+		var schedInformerFactory schedinformers.SharedInformerFactory
 		if namespace == metav1.NamespaceAll {
 			kubeInformerFactory = kubeinformers.NewSharedInformerFactory(kubeClient, 0)
 			kubeflowInformerFactory = informers.NewSharedInformerFactory(mpiJobClientSet, 0)
 			volcanoInformerFactory = volcanoinformers.NewSharedInformerFactory(volcanoClientSet, 0)
+			schedInformerFactory = schedinformers.NewSharedInformerFactory(schedClientSet, 0)
```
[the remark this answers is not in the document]
tenzen-y: That's a great idea.

go.mod
tenzen-y: Yes, the newest version is 0.25. 
Also, according to https://github.com/kubernetes-sigs/scheduler-plugins/issues/503, the scheduler-plugins seems to be released every 3 months.
So 0.26 will be released in June.

> Yes, we do. Usually, we cut official releases every ~3 months

manifests/base/cluster-role.yaml
```diff
@@ -89,6 +89,20 @@ rules:
...
   verbs:
   - "*"
+- apiGroups:
+  - scheduling.x-k8s.io
+  resources:
+  - podgroups
+  verbs:
+    - "*"
+- apiGroups:
+  - scheduling.k8s.io
```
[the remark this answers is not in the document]
tenzen-y: Does that mean the below?

```yaml
- apiGroups:
  - scheduling.x-k8s.io
  - scheduling.k8s.io
  resources:
  - podgroups
  - priorityclasses
  verbs:
    - "*"
```

pkg/controller/mpi_job_controller.go
tenzen-y: Oh, sorry. This is my bad.

pkg/controller/mpi_job_controller_test.go
tenzen-y: Sure.

cmd/mpi-operator/app/options/options.go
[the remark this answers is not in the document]
tenzen-y: Done.

cmd/mpi-operator/app/server.go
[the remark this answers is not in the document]
tenzen-y: Done.

pkg/controller/mpi_job_controller.go
tenzen-y: Done.

pkg/controller/mpi_job_controller_test.go
tenzen-y: Done.

manifests/base/cluster-role.yaml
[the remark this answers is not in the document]
alculquicondor: Sorry. I meant to ask if we really need to support both groups or just the newer one. I would vote for just the newer one.

pkg/controller/podgroup.go
alculquicondor: resource calculation in scheduler is: `max(initContainers, sum(containers)) + pod overhead`.

Also, this is ignoring any limit ranges or defaulting.

But I guess it's ok if this is just an estimate. I would guess that the scheduler ultimately considers the final pods.

manifests/base/cluster-role.yaml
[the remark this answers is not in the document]
tenzen-y: Note:
- OLD: scheduling.k8s.io/v1alpha1 PodGroup
- NEW: scheduling.x-k8s.io/v1alpha1 PodGroup

pkg/controller/podgroup.go
tenzen-y: > But I guess it's ok if this is just an estimate. I would guess that the scheduler ultimately considers the final pods.

I think so too. It's fine with only containers since the conscheduling plugin isn't quota management software :)

alculquicondor:
I wonder if @denkensk could give a sanity check as one of the authors of the coscheduling plugin.

tenzen-y:
> I wonder if @denkensk could give a sanity check as one of the authors of the coscheduling plugin.

@alculquicondor Thank you for the review! 
I'd be happy if this PR is reviewed by @denkensk.

pkg/controller/podgroup.go
```diff
@@ -58,19 +124,225 @@ func newPodGroup(mpiJob *kubeflow.MPIJob) *volcanov1beta1.PodGroup {
...
+		// NUM(Workers) = minMember - NUM(Launcher)
+		order[1].Replicas = pointer.Int32(*minMember - 1)
+	}
+
+	minResources := corev1.ResourceList{}
+	for _, rp := range order {
+		if rp.Replicas == nil {
+			continue
+		}
+		for i := int32(0); i < *rp.Replicas; i++ {
```
alculquicondor, line 321: Uhm... can you replace this with a multiplication?

alculquicondor:
/lgtm
/hold
for @denkensk

pkg/controller/podgroup.go
tenzen-y: Does that mean `for i := 0; i < (num(containers) × replicas); i++`?

alculquicondor: I mean, avoid iterating over and over the pod containers. You should be able to multiply the resources for 1 pod by the number of replicas

tenzen-y: I see. Makes sense. I will try.

tenzen-y: Done.

pkg/controller/podgroup.go
```diff
@@ -374,10 +374,11 @@ func addResources(minResources corev1.ResourceList, resources corev1.ResourceReq
 		}
 	}
 	for name, quantity := range merged {
+		addition := resource.MustParse(strconv.Itoa(int(quantity.Value()) * replicas))
```
alculquicondor, line 377: Ah, I didn't realize that Quantity doesn't have a multiplication operation.

This is an odd way of calculating it though. Have you verified if this works for something like `1m`?

If not, maybe we can revert and follow up with a PR upstream to add multiplication.

alculquicondor: FWIW, in kueue we convert everything to int64 before multiplying and assume, using `MilliValue` for CPU.

tenzen-y: > Have you verified if this works for something like 1m?

I verified that situation right now. This way doesn't work well :(

tenzen-y: > FWIW, in kueue we convert everything to int64 before multiplying, using MilliValue for CPU.

I verified  the `resource.MustParse(strconv.Itoa(int(quantity.MilliValue()) * replicas))` works well.

tenzen-y: However, as you say, this way looks odd.

alculquicondor: are you sure it gives you the right value?

also, it might not work well with memory.

tenzen-y: > are you sure it gives you the right value?
>
> also, it might not work well with memory.

If we use `MilliValue` for CPU and Memory, it seems to work well. However, for GPU, it doesn't work well.

tenzen-y: However, I’m not sure the above conversion works well in all situations.

So, It might be better to revert this change.

alculquicondor: +1

we should follow up with a multiplication method upstream. It will become increasingly useful for batch operators.

tenzen-y: I will create an issue in k/k tomorrow or next week.

tenzen-y: reverted.

pkg/controller/mpi_job_controller_test.go
```diff
@@ -148,36 +158,42 @@ func newMPIJob(name string, replicas *int32, startTime, completionTime *metav1.T
...
 	i := informers.NewSharedInformerFactory(f.client, noResyncPeriodFunc())
 	k8sI := kubeinformers.NewSharedInformerFactory(f.kubeClient, noResyncPeriodFunc())
 
-	volcanoInformerFactory := volcanoinformers.NewSharedInformerFactory(f.volcanoClient, 0)
-	podgroupsInformer := volcanoInformerFactory.Scheduling().V1beta1().PodGroups()
+	var podGroupCtrl PodGroupControl
+	if f.gangSchedulingName == options.GangSchedulerVolcano {
+		podGroupCtrl = NewVolcanoCtrl(f.volcanoClient, metav1.NamespaceAll)
+	} else if len(f.gangSchedulingName) != 0 {
+		podGroupCtrl = NewSchedulerPluginsCtrl(f.schedClient, metav1.NamespaceAll, "default-scheduler")
```
denkensk, line 171: is it “default-scheduler”?

pkg/controller/mpi_job_controller.go
```diff
@@ -213,23 +211,24 @@ type MPIJobController struct {
...
-	jobLister       batchlisters.JobLister
-	jobSynced       cache.InformerSynced
-	podLister       corelisters.PodLister
-	podSynced       cache.InformerSynced
-	podgroupsLister podgroupslists.PodGroupLister
-	podgroupsSynced cache.InformerSynced
-	mpiJobLister    listers.MPIJobLister
-	mpiJobSynced    cache.InformerSynced
+	// podGroupCtrl is a client for PodGroups (volcano and scheduler-plugins).
+	podGroupCtrl PodGroupControl
```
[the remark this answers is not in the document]
tenzen-y: I'm not the special reason for calling it "xxControl", and I was inspired by https://github.com/kubeflow/common/tree/83259a047fc47895dab1012728fb829523363a6d/pkg/controller.v1/control. This means we can even name it "xxClient".

cmd/mpi-operator/app/server.go
```diff
@@ -133,40 +133,44 @@ func Run(opt *options.ServerOption) error {
...
-		if opt.GangSchedulingName != "" {
-			podgroupsInformer = volcanoInformerFactory.Scheduling().V1beta1().PodGroups()
+		kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClient, 0, kubeInformerFactoryOpts...)
+		kubeflowInformerFactory := informers.NewSharedInformerFactoryWithOptions(mpiJobClientSet, 0, kubeflowInformerFactoryOpts...)
+
+		// For the gang scheduling
+		var podGroupCtrl controllersv1.PodGroupControl
+		if opt.GangSchedulingName == options.GangSchedulerVolcano {
+			podGroupCtrl = controllersv1.NewVolcanoCtrl(volcanoClientSet, namespace)
+		} else if len(opt.GangSchedulingName) != 0 {
```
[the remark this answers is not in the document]
tenzen-y: If you set some scheduler name excluding `volcano` to the gang-scheduling flag, the mpi-operator uses scheduler-plugins as we decided to block collaborating with some custom schedulers in https://github.com/kubeflow/common/pull/209#issuecomment-1462723903.

pkg/controller/mpi_job_controller_test.go
tenzen-y: IIUC, if we set up the scheduler-plugins as single scheduler mode, we can use the scheduler-plugins as a part of the default scheduler.

https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/doc/install.md#install-release-v0257-and-use-coscheduling

cmd/mpi-operator/app/server.go
```diff
@@ -259,29 +263,43 @@ func Run(opt *options.ServerOption) error {
...
+		return nil, nil, nil, nil, nil, err
 	}
 
 	volcanoClientSet, err := volcanoclient.NewForConfig(restclientset.AddUserAgent(config, "volcano"))
 	if err != nil {
-		return nil, nil, nil, nil, err
+		return nil, nil, nil, nil, nil, err
+	}
+
+	schedClientSet, err := schedclientset.NewForConfig(restclientset.AddUserAgent(config, "scheduler-plugins"))
```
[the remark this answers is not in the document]
tenzen-y: Ah, it might better pass the gangScheulingName to createClientSets func.

tenzen-y:
@alculquicondor Squashed.

alculquicondor:
/approve

leaving lgtm to @denkensk

denkensk:
/lgtm

Thanks for your contribution. @tenzen-y

PR #542, opened 2023-03-31
Update podgroups once schedulingPolicy of MPIJobs are changed

tenzen-y:
Previously, we implemented the logic to create and delete podGroups for the scheduler plugins in #538.

However, I forgot to implement the logic to update podGroups.

So, I implemented the logic. Also, I cleaned up the PodGroupCtrl interface so that we can control PodGroups without passing an unneeded namespace.

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -673,10 +673,11 @@ func (c *MPIJobController) getLauncherJob(mpiJob *kubeflow.MPIJob) (*batchv1.Job
 
 // getOrCreatePodGroups will create a PodGroup for gang scheduling by volcano.
 func (c *MPIJobController) getOrCreatePodGroups(mpiJob *kubeflow.MPIJob) (metav1.Object, error) {
-	podGroup, err := c.podGroupCtrl.getPodGroup(mpiJob.Namespace, mpiJob.Name)
+	newPodGroup := c.podGroupCtrl.newPodGroup(mpiJob)
+	podGroup, err := c.podGroupCtrl.getPodGroup(newPodGroup.GetNamespace(), newPodGroup.GetName())
 	// If the PodGroup doesn't exist, we'll create it.
 	if errors.IsNotFound(err) {
-		podGroup, err = c.podGroupCtrl.createPodGroup(context.TODO(), mpiJob.Namespace, c.podGroupCtrl.newPodGroup(mpiJob))
+		return c.podGroupCtrl.createPodGroup(context.TODO(), newPodGroup)
 	}
 	// If an error occurs during Get/Create, we'll requeue the item so we
 	// can attempt processing again later. This could have been caused by a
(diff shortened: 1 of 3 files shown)


alculquicondor:
do we have the same behavior in training-operator?

Question about coscheduling plugin:
I suppose that if the minMembers increases, but pods were already scheduled, they wouldn't be disrupted. But new pods will be blocked from scheduling until the new minMembers threshold is reached. Correct?

tenzen-y:
> do we have the same behavior in training-operator?

Yes, the training operator updates podGroups once schedulingPolicy is updated.

https://github.com/kubeflow/common/blob/83259a047fc47895dab1012728fb829523363a6d/pkg/controller.v1/common/scheduling.go#L49-L51

> But new pods will be blocked from scheduling until the new minMembers threshold is reached. Correct?

Do new pods mean the case of increased worker replicas of MPIJob (scale out)?

pkg/controller/mpi_job_controller.go
```diff
@@ -691,6 +692,10 @@ func (c *MPIJobController) getOrCreatePodGroups(mpiJob *kubeflow.MPIJob) (metav1
 		c.recorder.Event(mpiJob, corev1.EventTypeWarning, ErrResourceExists, msg)
 		return nil, fmt.Errorf(msg)
 	}
+
+	if !c.podGroupCtrl.pgSpecsAreEqual(podGroup, newPodGroup) {
```
alculquicondor, line 696: nit:
copy the new spec here. `podGroupCtrl` should just receive the updated object.

test/integration/mpi_job_controller_test.go
```diff
@@ -535,6 +536,23 @@ func TestMPIJobWithSchedulerPlugins(t *testing.T) {
...
+		t.Errorf("Failed sending job to apiserver: %v", err)
+	}
+	if err = wait.Poll(waitInterval, wait.ForeverTestTimeout, func() (bool, error) {
+		pg, err := getSchedPodGroup(ctx, gangSchedulerCfg.schedClient, mpiJob)
+		if err != nil {
+			return false, err
+		}
+		return *pg.Spec.ScheduleTimeoutSeconds == updatedScheduleTimeSeconds, nil
+	}); err != nil {
+		t.Errorf("Failed updating shceduler-plugins PodGroup: %v", err)
```
alculquicondor, line 554:
```suggestion
		t.Errorf("Failed updating scheduler-plugins PodGroup: %v", err)
```

tenzen-y:
> > Do new pods mean the case of increased worker replicas of MPIJob (scale out)?
> > yes

Correct.

test/integration/mpi_job_controller_test.go
tenzen-y: Oops

alculquicondor:
did you forget to push?

tenzen-y:
> did you forget to push?

@alculquicondor Sorry. I forgot to push.

pkg/controller/mpi_job_controller.go
```diff
@@ -691,6 +692,10 @@ func (c *MPIJobController) getOrCreatePodGroups(mpiJob *kubeflow.MPIJob) (metav1
 		c.recorder.Event(mpiJob, corev1.EventTypeWarning, ErrResourceExists, msg)
 		return nil, fmt.Errorf(msg)
 	}
+
+	if updatedPodGroup, equal := c.podGroupCtrl.pgSpecsAreEqual(podGroup, newPodGroup); !equal {
```
alculquicondor, line 696: oh, I see what you did and now I understand why the previous implementation was necessary: you can't update the spec unless you know the concrete type.

Sorry for the back and forth, but I think the previous implementation was better

tenzen-y: I'm ok. Let's revert this commit.

pkg/controller/mpi_job_controller.go
alculquicondor: as discussed, this is fine

alculquicondor (commented):
/lgtm

PR #543, opened 2023-04-04
Fix the logic to calculate minResources

tenzen-y:
I fixed the logic to calculate minResources so that `calculatePGMinResource` treats the launcher as a replica of higher priority when we don't set priorityClasses.

I faced the issue at https://github.com/kubeflow/mpi-operator/pull/540#issuecomment-1496012813.

Background: 
In the current implementation, if the launcher and workers have the same priority, `calculatePGMinResource` randomly selects prioritized replicas. This means the launcher might be treated as a lower priority than the worker replica when we don't set priorityClass in both replicas.


tenzen-y:
Rebased.

PR #550, opened 2023-04-15
(add) mpi_job_duration_histogram metric with linearBuckets

ghost:
Hello 👋  This PR adds a new Prometheus histogram metric vec to monitor the duration of MPI jobs, providing insights into the performance and efficiency of the MPI operator. New metric is named to mpi_operator_jobs_duration_seconds and covers a range of MPI job durations from 5 secs up to a month with 31 buckets.

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -975,6 +975,13 @@ func (c *MPIJobController) deleteWorkerPods(mpiJob *kubeflow.MPIJob) error {
 	return nil
 }
 
+func observeDuration(mpiJob *kubeflow.MPIJob, launcher *batchv1.Job, status string) {
+	if mpiJob.Status.StartTime != nil && mpiJob.Status.CompletionTime != nil {
+		duration := mpiJob.Status.CompletionTime.Sub(mpiJob.Status.StartTime.Time).Seconds()
+		mpiJobsDurationHistogram.WithLabelValues(launcher.Name, mpiJob.Namespace, status).Observe(duration)
+	}
+}
+
 func (c *MPIJobController) updateMPIJobStatus(mpiJob *kubeflow.MPIJob, launcher *batchv1.Job, worker []*corev1.Pod) error {
 	oldStatus := mpiJob.Status.DeepCopy()
 	if isMPIJobSuspended(mpiJob) {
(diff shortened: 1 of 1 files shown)


pkg/controller/mpi_job_controller.go
```diff
@@ -1004,9 +1009,20 @@ func (c *MPIJobController) updateMPIJobStatus(mpiJob *kubeflow.MPIJob, launcher
...
+				duration := mpiJob.Status.CompletionTime.Sub(mpiJob.Status.StartTime.Time).Seconds()
+				mpiJobsDurationHistogram.WithLabelValues(launcher.Name, mpiJob.Namespace, "status: job success").Observe(duration)
+			}
 		} else if isJobFailed(launcher) {
 			c.updateMPIJobFailedStatus(mpiJob, launcher, launcherPods)
+
+			if mpiJob.Status.StartTime != nil && mpiJob.Status.CompletionTime != nil {
+				duration := mpiJob.Status.CompletionTime.Sub(mpiJob.Status.StartTime.Time).Seconds()
+				mpiJobsDurationHistogram.WithLabelValues(launcher.Name, mpiJob.Namespace, "status: job failure").Observe(duration)
+			}
```
tenzen-y, line 1025: L1015-L1018 seems similar to L1022-L1025. Maybe, only the status message is different?
Can you commonize the process?

tenzen-y (commented):
@dogukanutuna Thank you for creating this PR!
I left a few comments.

pkg/controller/mpi_job_controller.go
```diff
@@ -134,6 +134,11 @@ var (
 		Name: "mpi_operator_job_info",
 		Help: "Information about MPIJob",
 	}, []string{"launcher", "namespace"})
+	mpiJobsDurationHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "mpi_operator_jobs_duration_seconds",
+		Help:    "Duration of MPI jobs in seconds",
+		Buckets: prometheus.LinearBuckets(5, 86400, 31), // 5s to 2,592,000 seconds  (a month)
```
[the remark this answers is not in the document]
alculquicondor: This is a big number of buckets, which I expect to be problematic in terms of memory. Are you sure we can't use Exponential? a base 2 is probably good enough.

tenzen-y (commented):
Thanks for the updates!
/lgtm

@alculquicondor Can you approve the CI?

pkg/controller/mpi_job_controller.go
```diff
@@ -134,6 +134,11 @@ var (
 		Name: "mpi_operator_job_info",
 		Help: "Information about MPIJob",
 	}, []string{"launcher", "namespace"})
+	mpiJobsDurationHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "mpi_operator_jobs_duration_seconds",
+		Help:    "Duration of MPI jobs in seconds",
+		Buckets: prometheus.LinearBuckets(0, 10, 60),
```
alculquicondor, line 140: 60 buckets of 10 seconds, for a total of 10min? Are you sure this is what's most useful?
In any case, leave a comment for the max bucket.

PR #523, opened 2023-02-09
Add slots to hostfile

tenzen-y:
Signed-off-by: Yuki Iwai <yuki.iwai.tz@gmail.com>

I added slots to the hostfile for the Horovod.

ref: https://horovod.readthedocs.io/en/stable/running_include.html#run-horovod

Fixes: #445

PR #562, opened 2023-06-08
MPICH support

sheevy:
This is meant to be a continuation of https://github.com/kubeflow/mpi-operator/pull/478 The original PR was stale, and master moved a lot, so it was easier to just create a new PR. I hope that is ok.

These are meant to be the same changes as in the https://github.com/kubeflow/mpi-operator/pull/478, but rebased on top of current master. The main problem with previous PR was the fact that SlotsPerWorker used enviroment variable to control number of slots, but unfortunately such variable does not exist in case for MPICH. Suggested solution was to add number of slots per worker to hostfile. This PR does not implement this, because it was already done in https://github.com/kubeflow/mpi-operator/pull/523

I hope that's correct understanding.


tenzen-y:
Thank you for creating this PR!
I will review this tomorrow.

tenzen-y:
@terrytangyuan Can you approve CI?

terrytangyuan:
Looks failed

tenzen-y:
@sheevy Can you address the error in CI?

sheevy:
Do you have any hints where to look? I couldn't see anything in the logs which would point in the right direction

tenzen-y:
> Do you have any hints where to look? I couldn't see anything in the logs which would point in the right direction

CI says we need to regenerate manifests. Did you run `make generate` on your local?

sheevy:
I have not. Thanks for the hint, I will check that tomorrow.

tenzen-y:
Also, you can reproduce error with `make verify-generate` on your local.

https://github.com/kubeflow/mpi-operator/blob/fda0532ba1a25e48b68316759c2dec438cfc8d13/Makefile#L89-L91

pkg/controller/mpi_job_controller.go
```diff
@@ -1203,6 +1214,8 @@ func newConfigMap(mpiJob *kubeflow.MPIJob, workerReplicas int32) *corev1.ConfigM
 			buffer.WriteString(fmt.Sprintf("%s%s-%d.%s.%s.svc slots=%d\n", mpiJob.Name, workerSuffix, i, workersService, mpiJob.Namespace, slots))
 		case kubeflow.MPIImplementationIntel:
 			buffer.WriteString(fmt.Sprintf("%s%s-%d.%s.%s.svc:%d\n", mpiJob.Name, workerSuffix, i, workersService, mpiJob.Namespace, slots))
+		case kubeflow.MPIImplementationMPICH:
+			buffer.WriteString(fmt.Sprintf("%s%s-%d.%s.%s.svc:%d\n", mpiJob.Name, workerSuffix, i, workersService, mpiJob.Namespace, slots))
```
tenzen-y, line 1218:
```suggestion
		case kubeflow.MPIImplementationIntel, kubeflow.MPIImplementationMPICH:
			buffer.WriteString(fmt.Sprintf("%s%s-%d.%s.%s.svc:%d\n", mpiJob.Name, workerSuffix, i, workersService, mpiJob.Namespace, slots))
```

build/base/mpich-entrypoint.sh
tenzen-y, line 1: This script seems mostly the same as the one for the intel mpi, right?
So, can we commonize?

Makefile
```diff
@@ -23,6 +23,7 @@ BASE_IMAGE_SSH_PORT?=2222
 IMG_BUILDER=docker
 PLATFORMS ?= linux/amd64
 INTEL_PLATFORMS ?= linux/amd64
+MPICH_PLATFORMS ?= linux/amd64
```
tenzen-y, line 26: Can we build MPICH container images for linux/arm64 and linux/ppc64le?

examples/v2beta1/pi/mpich.Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+FROM mpioperator/mpich-builder:${BASE_LABEL} as builder
+
+COPY pi.cc /src/pi.cc
+RUN mpic++ /src/pi.cc -o /pi
+
+
+FROM mpioperator/mpich:${BASE_LABEL}
+
+COPY --from=builder /pi /home/mpiuser/pi
```
tenzen-y, line 11: Adding an empty line would be good.

pkg/apis/kubeflow/validation/validation.go
```diff
@@ -35,7 +35,8 @@ var (
 
 	validMPIImplementations = sets.NewString(
 		string(kubeflow.MPIImplementationOpenMPI),
-		string(kubeflow.MPIImplementationIntel))
+		string(kubeflow.MPIImplementationIntel),
+		string(kubeflow.MPIImplementationMPICH))
```
tenzen-y, line 39: Should we update https://github.com/kubeflow/mpi-operator/blob/fda0532ba1a25e48b68316759c2dec438cfc8d13/pkg/apis/kubeflow/validation/validation_test.go as well?

test/e2e/mpi_job_test.go
```diff
@@ -217,6 +217,54 @@ var _ = ginkgo.Describe("MPIJob", func() {
 
 	})
 
+	ginkgo.Context("with MPICH Implementation", func() {
```
tenzen-y, line 220: Great test!

pkg/controller/mpi_job_controller.go
```diff
@@ -199,6 +199,16 @@ var (
...
 		},
 	}
+	mpichEnvVars = []corev1.EnvVar{
+		{
+			Name:  "HYDRA_HOST_FILE",
+			Value: fmt.Sprintf("%s/%s", configMountPath, hostfileName),
+		},
+		{
+			Name:  "HYDRA_LAUNCH_EXTRA_ARGS",
+			Value: "-o ConnectionAttempts=10",
```
tenzen-y, line 209: Why do we need to set this option? Wouldn't `build/base/mpich-entrypoint.sh` be enough?

tenzen-y: Or, we might be able to use only this option and remove `build/base/mpich-entrypoint.sh`.

wdyt?

pkg/controller/mpi_job_controller.go
sheevy: done

build/base/mpich-entrypoint.sh
sheevy: done. intel-entrypoint already checks for intel specific files annd only call the intel setup script if they are present. So I can use the same entrypoint.sh script for Intel and MPICH. So I just remove mpich-entrypoint.sh, renamed intel-etrypoint.sh to just entrypoint.sh and I use it in both cases

Makefile
sheevy: I have no experience with that, or a way of testing it. Other MPI implementations don't use it. So I would feel uneasy adding it without knowing it actually works.

pkg/controller/mpi_job_controller.go
sheevy: Keep in mind I'm not an exper in eithere mpi-operator or MPI(CH), but I think:
1. we need entrypoint.sh to prepare the hostfile, so it needs to stay.
2. IntelMPI is based on MPICH, so I've just copied HYDRA_LAUNCH_EXTRA_ARGS as equivalant to Intel's I_MPI_HYDRA_BOOTSTRAP_EXEC_EXTRA_ARGS few lines above.
3. I would have guessed that the `max_retry=10` controls number of SSH connection attempts, and only after SHH connection is established, there are 10 connection attempts on Hydra/MPI side. So I think those params have different meaning.

Makefile
tenzen-y: In OpenMPI, we are providing container images for linux/arm64 and linux/ppele64. here: https://hub.docker.com/layers/mpioperator/openmpi-builder/v0.4.0/images/sha256-e193e5365a2777d33d964f5563f4e5bdb7c7d1e5351c46dfbf7a436ea9338633?context=explore

So, supporting their platforms would be good.

tenzen-y: If you don't have environments for testing them, creating an issue to keep tracking them would be well.

pkg/controller/mpi_job_controller.go
tenzen-y: Thanks for the great clarification! I agree with you. Let's keep using both of them.

Makefile
sheevy: Sure. I won;t be adding support for those architectures, I I've created the issue, as suggested: https://github.com/kubeflow/mpi-operator/issues/565

pkg/apis/kubeflow/validation/validation.go
tenzen-y: I'm yet waiting for this.

tenzen-y:
Also, can you update the PR title with `MPICH support`?

tenzen-y (commented):
Overall LGTM excepted a unit test for validation.

/assign @alculquicondor

pkg/apis/kubeflow/validation/validation.go
sheevy: Should be in now. I tried to refactor into a common function, so both Intel and MPICH tests just call it. As the part of that refactoring "valid with worker" tests uses has different restart policy, then it had before. However, I'm fine with that as it's not what's being tested.

sheevy:
I think I've implemented all the comments and suggestions. I'm happy for it to get re-tested. Please let me know if you have any further feedback.

sheevy:
Hey @terrytangyuan, can you approve another run, fixes for all suggestions are in?

pkg/apis/kubeflow/validation/validation.go
tenzen-y: Thanks. Let me check.

pkg/apis/kubeflow/validation/validation_test.go
```diff
@@ -26,71 +26,66 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
+func GenerateValidJob(mpiImplementation kubeflow.MPIImplementation, hasWorkers bool) kubeflow.MPIJob {
+
```
tenzen-y, line 30: Generally, we don't prefer to generate test data in a unit test programmably. If we introduce programmable test data, I prefer to use method chains.

@alculquicondor WDYT?

sheevy: Sure, I will adjust the code once there's a recommendation.
My goal was to reduce repetition, to generate a valid test for both Intel and MPICH using the same code. Is it possible to achieve using method chains? Or shall I just have a copy for each MPI implementation?

tenzen-y: I imagined things like kueue: https://github.com/kubernetes-sigs/kueue/blob/bd3caa823f0318d636ac39241938da987b31a9bf/pkg/util/testing/wrappers.go

However, wrappers may be overkill for the mpi-operator.
Let me know what @alculquicondor thinks.

terrytangyuan:
Sure

build/base/entrypoint.sh
alculquicondor, line 1: Was this necessary?

In OpenMPI, we don't need this file, because OpenMPI is failure tolerant and implements retries when a host is not found.

IntelMPI just fails. What happens with MPICH?

build/base/mpich-builder.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM debian:bullseye as builder
```
alculquicondor, line 1: Ideally, we should be using a newer version, but given that `mpioperator/base` still uses bullseye, this makes sense.

If you have the chance, I would welcome a PR to update the base image everywhere.

examples/v2beta1/pi/mpich.Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG BASE_LABEL
+
+FROM mpioperator/mpich-builder:${BASE_LABEL} as builder
+
+COPY pi.cc /src/pi.cc
```
alculquicondor, line 5: glad to see that the same code works for every implementation 😄

pkg/apis/kubeflow/validation/validation_test.go
alculquicondor: -1 on this function.

https://enterprisecraftsmanship.com/posts/dry-damp-unit-tests/

We also don't need wrappers for so few cases. Just be expressive.

pkg/controller/mpi_job_controller.go
alculquicondor: These arguments affect the TCP+TLS connection, but they don't affect the DNS resolution, IIRC. So we might need both this and the entrypoint. But if we can get rid of the entry point, that would be better. Worth experimenting.

alculquicondor: > we need entrypoint.sh to prepare the hostfile, so it needs to stay.

The entry point doesn't prepare the hostfile. It just checks that all hosts listed in the hostfile are resolvable.

test/e2e/mpi_job_test.go
```diff
@@ -217,6 +217,54 @@ var _ = ginkgo.Describe("MPIJob", func() {
...
+		ginkgo.When("running as root", func() {
```
alculquicondor, line 221: Can we also have another test for running as user?

build/base/entrypoint.sh
sheevy: I recall having tested it last year, during the previous PR. And it was required. This makes sense, since IntelMPI is based on MPICH.

build/base/mpich-builder.Dockerfile
sheevy: Noted. Happy to do it in a separate PR.

alculquicondor (commented):
/lgtm
/approve

Thanks for adding the test for Intel as well!

Issue #570, opened 2023-06-22
e2e test failed sometime


lowang-bh:
https://github.com/kubeflow/mpi-operator/actions/runs/5341853866/jobs/9683080365?pr=569

```
• [SLOW TEST:48.021 seconds]
MPIJob
/home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:39
  with OpenMPI implementation
  /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:85
    when running as root
    /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:100
      when running with host network
      /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:130
        should succeed
        /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:140
------------------------------
W0622 05:27:11.815081   31530 warnings.go:70] unknown field "spec.mpiReplicaSpecs.Launcher.template.metadata.creationTimestamp"
W0622 05:27:11.815092   31530 warnings.go:70] unknown field "spec.mpiReplicaSpecs.Worker.template.metadata.creationTimestamp"
• [SLOW TEST:53.020 seconds]
MPIJob
/home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:39
  with OpenMPI implementation
  /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:85
    when running as non-root
    /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:147
      should succeed
      /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:164
------------------------------
W0622 05:28:04.845095   31530 warnings.go:70] unknown field "spec.mpiReplicaSpecs.Launcher.template.metadata.creationTimestamp"
W0622 05:28:04.845129   31530 warnings.go:70] unknown field "spec.mpiReplicaSpecs.Worker.template.metadata.creationTimestamp"
• [SLOW TEST:60.029 seconds]
MPIJob
/home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:39
  with Intel Implementation
  /home/runner/work/mpi-operator/mpi-operator/test/e2e/mpi_job_test.go:172
    when running as root
[...]

PR #405, opened 2021-08-17
Increase E2E wait timeout

alculquicondor:
To reduce flakiness

PR #576, opened 2023-06-23
Increase the timeout for E2E tests

sheevy:
Similar to https://github.com/kubeflow/mpi-operator/pull/405


alculquicondor:
I don't think individual tests are timing out. It's the whole suite that is timing out.
Note that the default timeout for `go test` is 10 minutes, as documented here: https://pkg.go.dev/cmd/go#hdr-Testing_flags:~:text=sequence%2C%20if%20any.-,%2Dtimeout%20d,-If%20a%20test

You need to increase it in the Makefile

sheevy:
Thanks. I've doubled that timeout to 20 minutes

alculquicondor:
could you squash?

sheevy:
ok, I failed trying to squash. I'll try once again. Is there a way to cancel the tests so I don't waste resources?

sheevy:
ok, now squashing worked as intended

tenzen-y (commented):
Thanks!
/lgtm
/assign @alculquicondor

tenzen-y:
Fix: #570

PR #579, opened 2023-07-07
add custom setup.py to install mpijob module

### LINKED ISSUE #578: python setup.py doesn't appear to install?
The current setup.py doesn't have logic so that it installs anywhere? 

https://github.com/kubeflow/mpi-operator/blob/master/sdk/python/v2beta1/setup.py

And the name seems to be calling it kubeflow but the module is mpijob? E.g., you can import from the local directory because that is by default on the python path, but you can't do a setup.py install and go somewhere else and do the import. Has anyone else run into this?

vsoch:
Problem: we cannot currently install the python mpijob module
Solution: add a setup.py proper that is ignored by the generator

Notes:

 - I chose kubeflow-mpijob to reflect the name of the Python module (typical practice, with a prefix)
 - Tested with `pip install .`
 - I didn't test with the generator yet, but saw this pattern to add the file to be ignored that (hopefully) will work.

Future improvements:

- more documentation about install
- full examples directory
- release package to pypi

This will close #578


sdk/python/v2beta1/setup.py
```diff
@@ -9,8 +9,61 @@
...
+
+def setup():
+    """
+    Main entrypoint to run setup.
+    """
+    setuptools.setup(
+        name="kubeflow-mpijob",
+        version="0.0.1",
+        author="Kubeflow Authors",
+        author_email="wg-batch@kubernetes.io",
```
tenzen-y, line 35:
```suggestion
        author_email="kubeflow-discuss@googlegroups.com",
```

Kubernetes WG batch doesn't maintain mpi-operator. So we should set another one to here.
@terrytangyuan @alculquicondor What do you think about using `kubeflow-discuss@googlegroups.com`?

tenzen-y (commented):
@vandanavk Thanks for the great contribution!

sdk/python/v2beta1/setup.py
terrytangyuan: It’s better not to spam that mailing list with bug reports

vsoch: I usually put a faux GitHub email here - the `<name>@users.noreply.github.com` but if someone has an actual email, that is likely ideal. I'm not sure how often these emails get used, but I would say if the group is protected by someone needing to join, that would get around bots that scrape the emails (and spam) and maybe even be a good approach? Someone that wants help with the package should maybe reach out to this list?

vsoch: Okay I updated to `kubeflow@users.noreply.github.com`. it is definitely useless and fake, but will satisfy the pypi requirements if you ever decide to release there (you just need the contact field defined with something resembling an email).

terrytangyuan: No objection.

tenzen-y: sgtm

terrytangyuan (approved):
Thanks!

/lgtm

cliu-nuro:
Hi @vsoch , what's the timeline like to release this package to pypi?

vsoch:
hey @cliu-nuro - there is no timeline, it's more that we could do it when it's requested. It isn't super hard, a maintainer/owner would need to create the first (dummy) deployment locally, usually you just do:

```bash
python setup.py sdist
twine upload dist/<path-to-tar-gz>
```
or create the package manually in the UI. Then get a pypi token / username for the package, explicitly, and then you just need
a workflow like this:

```yaml
name: release mpijob

on:
  release:
    types: [created]

jobs:
  deploy:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v3

    - name: Install
      run: conda create --quiet --name mpijob twine

    - name: Install dependencies
      run: |
        export PATH="/usr/share/miniconda/bin:$PATH"
        source activate mpijob
        cd /path/to/where/mpijob/setup.py/is
        pip install -e .[all]
        pip install setuptools wheel twine
    - name: Build and publish
      env:
        TWINE_USERNAME: ${{ secrets.PYPI_USER }}
        TWINE_PASSWORD: ${{ secrets.PYPI_PASS }}
      run: |
        export PATH="/usr/share/miniconda/bin:$PATH"
        source activate mpijob
        python setup.py sdist bdist_wheel
        twine upload dist/*
```
And that should upload on releases. I'm not a maintainer here so I can't do most of that, but happy to provide guidance!

alculquicondor:
@terrytangyuan what was the process to release to pypi?

terrytangyuan:
I don't think we have a process yet. It's not published to PyPI.

terrytangyuan:
We can instruct users to install directly from GitHub.

Issue #586, opened 2023-08-11
[feature]upgrade volcano to v1.8.0


lowang-bh:
Volcano is about to release v1.8.0 recently. After that, it need to upgrade volcano dependency to 1.8.0

PR #589, opened 2023-08-25
update volcano scheduler to 1.8.0; ignore vendor

lowang-bh:
Fix issue #586
To keep update for volcano release 1.8.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ profile.cov
 hack/python-sdk/openapi-generator-cli.jar
 dep-crds/
 dep-manifests/
+vendor/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ profile.cov
 hack/python-sdk/openapi-generator-cli.jar
 dep-crds/
 dep-manifests/
+vendor/
```
alculquicondor, line 9: why is this necessary?

lowang-bh: In case of git add . to include the vendor dir.

tenzen-y: @lowang-bh We don't use vendoring in this repository. So I don't think we need this ignoring. Are there any reasons for this ignore?

lowang-bh: @tenzen-y sometimes user runs `go mod vendor`.  In case of `git add .` to include it. I can remove it if it is really not need.

tenzen-y: I'm ok with adding `vendor` to `.gitignore`.

alculquicondor (commented):
/approve

PR #600, opened 2023-11-04
Fix invalid link for horovod cpu-only example

lianghao208:
Fix invalid link for horovod cpu-only example.


alculquicondor (commented):
/lgtm
/approve

PR #601, opened 2023-11-04
Fix invalid link for horovod cpu-only example Dockerfile

lianghao208:
I just realized the link of Dockerfile is also invalid, submit another PR to fix.


terrytangyuan (commented):
Thanks!

/lgtm

PR #610, opened 2023-12-29
add deepspeed example

kuizhiqing:


examples/v2beta1/deepspeed/deepspeed-helloworld.yaml
```diff
@@ -0,0 +1,32 @@
...
+  slotsPerWorker: 1
+  runPolicy:
+    cleanPodPolicy: Running
+  mpiReplicaSpecs:
+    Launcher:
+      replicas: 1
+      template:
+        spec:
+          containers:
+          - image: registry.cn-beijing.aliyuncs.com/acs/deepspeed:hello-deepspeed
```
terrytangyuan, line 15: It would be useful to check in the Dockerfile as well.

kuizhiqing: You'er right. This demo is intend to be used to test the up coming feature, I'll manage it later.

vsoch:
@kuizhiqing do you have a suggested place to run / test a setup like this? I've been trying to get just 3 nodes each with a single GPU on Google Cloud and I never get the allocation. I was able to get a single node with one GPU last week but it felt like luck. Will it work without GPU?

vsoch:
It looks like it defaults to CPU, but it's not clear to me how communication is setup. Is it just using a shared volume at `/workspace`? if that's the case, what's the point of an operator that supports MPI?

kuizhiqing:
Hi @vsoch , actually, I've test it in the production environment of my affiliation. I've no idea where to run it in an open platform.

The workspace path of the example contains materials from https://github.com/microsoft/DeepSpeedExamples/tree/master/training/HelloDeepSpeed, it do not involve the communication process. The communication setup by pdsh with the hostfile provided by the mpi-operator.

For the CPU version, I'm afraid I cannot provide more information since I did not work with it.

vsoch:
>  https://github.com/microsoft/DeepSpeedExamples/tree/master/training/HelloDeepSpeed, it do not involve the communication process. The communication setup by pdsh with the hostfile provided by the mpi-operator.

There would need to be some variant of an MPI run in there, and the communication (with the MPI operator) would happen via ssh bootstrap. I'm trying to understand where that logic is here.

PR #617, opened 2024-01-26
fix the condition

### LINKED ISSUE #615: The operator still creates the launcher when launcherCreationPolicy is "WaitForWorkersReady" and  suspend is "true"
The operator still creates the launcher when `spec.launcherCreationPolicy is "WaitForWorkersReady"` and `spec.runPolicy.suspend is "true"`.

wang-mask:
fixes #615 

The original logic is:
```go
if launcher == nil {
	if mpiJob.Spec.LauncherCreationPolicy == kubeflow.LauncherCreationPolicyAtStartup ||  c.countReadyWorkerPods(worker) == len(worker) {
		launcher, err = c.kubeClient.BatchV1().Jobs(namespace).Create(context.TODO(), c.newLauncherJob(mpiJob), metav1.CreateOptions{})
		....
}
```
If the MPIJob is `suspended`, the `len(worker)` and `c.countReadyWorkerPods(worker)` would be both `0`. Then this judgment condition will be meet, and the launcher pod will be created.

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -639,7 +639,7 @@ func (c *MPIJobController) syncHandler(key string) error {
 			}
 		}
 		if launcher == nil {
-			if mpiJob.Spec.LauncherCreationPolicy == kubeflow.LauncherCreationPolicyAtStartup || c.countReadyWorkerPods(worker) == len(worker) {
+			if mpiJob.Spec.LauncherCreationPolicy == kubeflow.LauncherCreationPolicyAtStartup || (!isMPIJobSuspended(mpiJob) && c.countReadyWorkerPods(worker) == len(worker)) {
 				launcher, err = c.kubeClient.BatchV1().Jobs(namespace).Create(context.TODO(), c.newLauncherJob(mpiJob), metav1.CreateOptions{})
 				if err != nil {
 					c.recorder.Eventf(mpiJob, corev1.EventTypeWarning, mpiJobFailedReason, "launcher pod created failed: %v", err)


pkg/controller/mpi_job_controller.go
```diff
@@ -639,7 +639,7 @@ func (c *MPIJobController) syncHandler(key string) error {
 			}
 		}
 		if launcher == nil {
-			if mpiJob.Spec.LauncherCreationPolicy == kubeflow.LauncherCreationPolicyAtStartup || c.countReadyWorkerPods(worker) == len(worker) {
+			if mpiJob.Spec.LauncherCreationPolicy == kubeflow.LauncherCreationPolicyAtStartup || (!isMPIJobSuspended(mpiJob) && c.countReadyWorkerPods(worker) == len(worker)) {
```
alculquicondor, line 642: This looks wrong too. If the job is suspended, we shouldn't create the launcher, even if the policy is `AtStartup`.

cc @tenzen-y 

I think this issue has been there for a while?

alculquicondor: Oh, I see, we still create the job, but suspend it https://github.com/kubeflow/mpi-operator/blob/edbde8caeb09c4d778761c8a1b1307474f832bef/pkg/controller/mpi_job_controller.go#L1445

alculquicondor:
> Then this judgment condition will be meet, and the launcher pod will be created.

But we only create the launcher Job suspended.

Are you saying that the problem arises when you unsuspend the job and suddenly all the pods start?

Not sure if it's worth always creating the Job and only unsuspending if the workers are ready.

wang-mask:
I think the launcher pod should be created after all the workers are ready when the `spec.launcherCreationPolicy` is set to `"WaitForWorkersReady"`,  even if the MPIJob is suspend.
The current logic is not very consistent with `"WaitForWorkersReady"`.

pkg/controller/mpi_job_controller.go
tenzen-y: > Oh, I see, we still create the job, but suspend it
> 
> https://github.com/kubeflow/mpi-operator/blob/edbde8caeb09c4d778761c8a1b1307474f832bef/pkg/controller/mpi_job_controller.go#L1445

Yes, so I think such an issue doesn't exist.

tenzen-y:
> I think the launcher pod should be created after all the workers are ready when the spec.launcherCreationPolicy is set to "WaitForWorkersReady", even if the MPIJob is suspend.

I think that when the MPIJob is suspended, workers aren't created by the mpi-operator.
Could you clarify such a situation?

wang-mask:
> > I think the launcher pod should be created after all the workers are ready when the spec.launcherCreationPolicy is set to "WaitForWorkersReady", even if the MPIJob is suspend.
> 
> I think that when the MPIJob is suspended, workers aren't created by the mpi-operator. Could you clarify such a situation?



> > I think the launcher pod should be created after all the workers are ready when the spec.launcherCreationPolicy is set to "WaitForWorkersReady", even if the MPIJob is suspend.
> 
> I think that when the MPIJob is suspended, workers aren't created by the mpi-operator. Could you clarify such a situation?

In this case the workers are not created by the mpi-operator, but the  mpi-operator creates the launcher when the spec.launcherCreationPolicy is set to "WaitForWorkersReady". 
Shouldn't it create the launcher after the mpi job is unsuspended and workers are ready?

tenzen-y:
> 
> 
> > I think that when the MPIJob is suspended, workers aren't created by the mpi-operator. Could you clarify such a situation?
> 
> 
> 
> In this case the workers are not created by the mpi-operator, but the  mpi-operator creates the launcher when the spec.launcherCreationPolicy is set to "WaitForWorkersReady". 
> 
> Shouldn't it create the launcher after the mpi job is unsuspended and workers are ready?

In this situation, who does create the workers?

wang-mask:
> > > I think that when the MPIJob is suspended, workers aren't created by the mpi-operator. Could you clarify such a situation?
> > 
> > 
> > In this case the workers are not created by the mpi-operator, but the  mpi-operator creates the launcher when the spec.launcherCreationPolicy is set to "WaitForWorkersReady".
> > Shouldn't it create the launcher after the mpi job is unsuspended and workers are ready?
> 
> In this situation, who does create the workers?

The operation of unsuspending an MPI job triggers the creation of workers.

alculquicondor:
With this implementation: what happens if the job is running, then it is suspended and unsuspended?

Is a launcher pod created as soon as it is unsuspended the second time? If so, this solution is not sufficient.

wang-mask:
> With this implementation: what happens if the job is running, then it is suspended and unsuspended?
> 
> Is a launcher pod created as soon as it is unsuspended the second time? If so, this solution is not sufficient.

Now the controller deletes the worker pod and suspends the launcher when the mpi job is suspended. 

Maybe it can also deletes the launcher when the mpi job is suspended and `spec.launcherCreationPolicy = "WaitForWorkersReady"`. After second or subsequent suspension was lifted, the controller waits for all workers to be ready before creating a launcher.

If this implementation is reasonable, I would be happy to modify this PR.

PR #618, opened 2024-01-30
adding timeout for cache sync

emsixteeen:
Currently, if informer cache syncing fails (typically because lack of permissions), there is no indication of this error. A message is logged indefinitely, the Pod remains in Running state, and "seems" to be healthy.

The PR adds a timeout to cache syncing, which will fail the controller if caches are not synced [within the timeout period].

### DIFF
--- a/cmd/mpi-operator/app/options/options.go
+++ b/cmd/mpi-operator/app/options/options.go
@@ -39,7 +39,7 @@ type ServerOption struct {
 	LockNamespace      string
 	QPS                int
 	Burst              int
-	CacheTimeout       time.Duration
+	CacheSyncTimeout   time.Duration
 }
 
 // NewServerOption creates a new CMServer with a default config.
@@ -78,6 +78,6 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 	fs.IntVar(&s.QPS, "kube-api-qps", 5, "QPS indicates the maximum QPS to the master from this client.")
 	fs.IntVar(&s.Burst, "kube-api-burst", 10, "Maximum burst for throttle.")
 
-	fs.DurationVar(&s.CacheTimeout, "cache-timeout", 30*time.Second,
+	fs.DurationVar(&s.CacheSyncTimeout, "cache-sync-timeout", 2*time.Minute,
 		`The amount of time to wait for caches to populate. If this timeout is exceeded, the mpi-operator will fail and exit.`)
 }
--- a/cmd/mpi-operator/app/options/options.go
+++ b/cmd/mpi-operator/app/options/options.go
@@ -17,6 +17,7 @@ package options
 import (
 	"flag"
 	"os"
+	"time"
 
 	"github.com/kubeflow/mpi-operator/pkg/apis/kubeflow/v2beta1"
 )
@@ -38,6 +39,7 @@ type ServerOption struct {
 	LockNamespace      string
 	QPS                int
 	Burst              int
+	CacheTimeout       time.Duration
 }
 
 // NewServerOption creates a new CMServer with a default config.
@@ -75,4 +77,7 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 
 	fs.IntVar(&s.QPS, "kube-api-qps", 5, "QPS indicates the maximum QPS to the master from this client.")
 	fs.IntVar(&s.Burst, "kube-api-burst", 10, "Maximum burst for throttle.")
+
+	fs.DurationVar(&s.CacheTimeout, "cache-timeout", 30*time.Second,
+		`The amount of time to wait for caches to populate. If this timeout is exceeded, the mpi-operator will fail and exit.`)
 }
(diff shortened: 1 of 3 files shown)


pkg/controller/mpi_job_controller.go
```diff
@@ -441,9 +441,22 @@ func (c *MPIJobController) Run(threadiness int, stopCh <-chan struct{}) error {
 	if c.PodGroupCtrl != nil {
 		synced = append(synced, c.podGroupSynced, c.priorityClassSynced)
 	}
-	if ok := cache.WaitForCacheSync(stopCh, synced...); !ok {
+
+	timeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
alculquicondor, line 445: What if you have so many jobs that 30s are not enough for the cache to sync?

emsixteeen: Changed the timeout duration to an option / flag. 

Not sure if that truly addresses the underlying question - but at least it's now configurable ...

cmd/mpi-operator/app/options/options.go
```diff
@@ -75,4 +77,7 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 
 	fs.IntVar(&s.QPS, "kube-api-qps", 5, "QPS indicates the maximum QPS to the master from this client.")
 	fs.IntVar(&s.Burst, "kube-api-burst", 10, "Maximum burst for throttle.")
+
+	fs.DurationVar(&s.CacheTimeout, "cache-timeout", 30*time.Second,
```
alculquicondor, line 81:
```suggestion
	fs.DurationVar(&s.CacheSyncTimeout, "cache-sync-timeout", 2*time.Minute,
```

alculquicondor (commented):
WDYT @tenzen-y?
Maybe a better approach would be to have the readiness probe be aware of whether the reconcilers are running?

tenzen-y:
> WDYT @tenzen-y? Maybe a better approach would be to have the readiness probe be aware of whether the reconcilers are running?

It makes sense. As I explore the timeout setting in the kube-controller-manager and the job-controller, the kube-controller-manager seems not to have such a setting for syncing cache and doesn't generate context with timeout.

alculquicondor:
>  It makes sense.

Sorry, can you clarify what exactly makes sense for you?

cmd/mpi-operator/app/options/options.go
emsixteeen: Updated / pushed per comments.

emsixteeen:
The use case at hand is to fail the controller early if access is denied to resources. Doing this via a timeout might be a bit of a kludge, as it doesn't really address the underlying issue.

After some more digging, it looks like an error handler can be set on informers: emsixteeen/mpi-operator@400b9fff.

Perhaps this is a better approach?

alculquicondor:
I like that approach better, but is there a way to do it just by iterating overall all in the informers that are instantiated for the SharedInformer?

tenzen-y:
> > It makes sense.
> 
> Sorry, can you clarify what exactly makes sense for you?

I meant that the kube-contoller-manager doesn't set such a timeout mechanism in the informer.
So, I would like to take readinessprobe approach.

However, setting the error handler to informers would be better than the readiness probe.

emsixteeen:
> I like that approach better, but is there a way to do it just by iterating overall all in the informers that are instantiated for the SharedInformer?

Not sure if it's possible to iterate over all the informers.

It seems that every _Informer_ returns its own interface, which usually has a method called `Informer()` which in turns returns the `SharedInformer` interface (via the `SharedIndexInformer` interface). It's the `SharedInformer` interface that provides the `SetWatchErrorHandler` method. 

I.e. I'm not aware of some sort of collection of `SharedInformer` interfaces that can easily be iterated over... 😕

alculquicondor:
Yeah, I don't think they are exposed in the informer factory.

alculquicondor:
Please update the PR with the error handler

emsixteeen:
In lieu of adding the error handler to this PR, I created a separate PR for it (#619). 

My thought process is that these PRs address two separate "features": this PR adds a timeout / limit for how long the operator will wait for caches to sync, and the other PR will cause the operator to fail if there are permission issues.

To @tenzen-y point, all of this could technically be handled in a readiness (or startup) probe, and Kubernetes could just fail the pod if it's unready after some time. However, the current readiness probe limits itself to leader election and is defined in [`server.go`](https://github.com/kubeflow/mpi-operator/blob/master/cmd/mpi-operator/app/server.go#L187), whereas these two "features" are handled in [`mpi_job_controller.go`](https://github.com/kubeflow/mpi-operator/blob/master/pkg/controller/mpi_job_controller.go). 

Because [`InstallPathHandler`](https://github.com/kubeflow/mpi-operator/blob/master/cmd/mpi-operator/app/server.go#L190) can only be [called once](https://github.com/kubernetes/apiserver/blob/v0.27.4/pkg/server/healthz/healthz.go#L159), some deeper refactoring would be needed to move these two health checks near each other, so as to ensure that `InstallPathHandler` is only called once.

These two "features" are probably valid on their own, regardless of the readiness probe... (?)

alculquicondor:
Assuming there are no authorization errors, I don't see a reason why to have a timeout for cache sync.

It should eventually finish.

/close

emsixteeen:
Fair enough! This was an initial attempt to address an issue with the wrong implementation. :blush:

PR #619, opened 2024-02-02
Fails mpi-operator early if access to list or watch objects is denied 

emsixteeen:
Currently the `mpi-operator` waits for informer caches to sync before being ready to do work.

If access to any objects fail due to permission issues, the failure is simply logged, and until or unless the access issues is resolved, the `mpi-operator` remains in this waiting state indefinitely.

This request adds an option to fatally fail the `mpi-operator` if the operator is unable to list or watch the objects due to Forbidden or Unauthorized errors.

The option defaults to `false` so that existing functionality / behavior is maintained.

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -346,6 +346,34 @@ func NewMPIJobControllerWithClock(
 
 	controller.updateStatusHandler = controller.doUpdateJobStatus
 
+	// Set up error handlers for informers
+	klog.Info("Setting up informer error handlers")
+	informers := map[string]cache.SharedInformer{
+		"configMapInformer":     configMapInformer.Informer(),
+		"secretInformer":        secretInformer.Informer(),
+		"serviceInformer":       serviceInformer.Informer(),
+		"jobInformer":           jobInformer.Informer(),
+		"podInformer":           podInformer.Informer(),
+		"priorityClassInformer": priorityClassInformer.Informer(),
+		"mpiJobInformer":        mpiJobInformer.Informer(),
+	}
+
+	for name, informer := range informers {
+		err := informer.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
+			// Pipe to default handler first, which just logs the error
+			cache.DefaultWatchErrorHandler(r, err)
+
+			if errors.IsUnauthorized(err) || errors.IsForbidden(err) {
+				klog.Fatalf("Unable to sync cache for informer %s: %s. Requesting controller to exit.", name, err)
+			}
+		})
+
+		if err != nil {
+			// return NewMPIJobControllerWithClock(...) (nil, error)
+			return nil, fmt.Errorf("unable to set error handler for informer %s: %s", name, err)
+		}
+	}
+
 	klog.Info("Setting up event handlers")
 	// Set up an event handler for when MPIJob resources change.
 	if _, err := mpiJobInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{


emsixteeen:
The thought process for making it an option was to have the `mpi-operator` fail fast (upon the first permission error).

Right now the [`MPIJobController`](/kubeflow/mpi-operator/blob/master/pkg/controller/mpi_job_controller.go#L282) startup process is something like this:

* Operator starts up, and [spawns Goroutines](/kubeflow/mpi-operator/blob/master/cmd/mpi-operator/app/server.go#L161-L165) to list/watch objects (`ConfigMaps`, `Secrets`, `Services`, etc.)
* It [waits](/kubeflow/mpi-operator/blob/master/pkg/controller/mpi_job_controller.go#L444) for `SharedInformer.HasSynced()` to be `true`
* Until the `SharedInformer.HasSynced()` returns `true`, the `mpi-operator` is in "waiting" state

Waiting for`SharedInformer.HasSynced()`to return `true` might be acceptable in some situations, such as if the permission issue is transient. 

Making the "fail-fast" optional (and defaulting it to `false`) just leaves the current behavior as is.

The use this request is trying to address:  when it'll be known upfront that the permission errors are  **not** transient, and you want the `mpi-controller` to fail fast.

Regarding _kubelet_ or the _Deployment Controller_ handling this, it seems that they would only handle this based on a readiness probe, which currently – once the leader is elected – will always come back as [healthy](/kubeflow/mpi-operator/blob/master/cmd/mpi-operator/app/server.go#L190). 

Per my [other comment](618#issuecomment-1924477153), all of this could technically be handled in a more robust readiness/health probe, but that poses other challenges (namely that [`InstallPathHandler`](/kubeflow/mpi-operator/blob/master/cmd/mpi-operator/app/server.go#L190) can only be [called once](/kubernetes/apiserver/blob/v0.27.4/pkg/server/healthz/healthz.go#L159)).

alculquicondor:
Regarding kubelet or the Deployment Controller handling this, it seems that they would only handle this based on a readiness probe, which currently – once the leader is elected – will always come back as [healthy](https://github.com/kubeflow/mpi-operator/blob/master/cmd/mpi-operator/app/server.go#L190).


What I mean is that if you enable fail-fast, we would have the kubelet restart the container or deployment controller to recreate the pod. So effectively, the E2E behavior is that the system continues trying to start the manager. And that E2E behavior is pretty similar to the existing behavior, except that it retries without exiting the binary.

Users are not losing anything. Instead, they are getting more visibility into any possible failures, as they will see pods being recreated or restarted, causing them to investigate.

pkg/controller/mpi_job_controller.go
```diff
@@ -346,6 +346,35 @@ func NewMPIJobControllerWithClock(
...
+			"mpiJobInformer":        mpiJobInformer.Informer(),
+		}
+
+		for name, informer := range informers {
+			err := informer.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
+				// Pipe to default handler first, which just logs the error
+				cache.DefaultWatchErrorHandler(r, err)
+
+				if errors.IsUnauthorized(err) || errors.IsForbidden(err) {
+					klog.Fatalf("Unable to sync cache for informer %s: %s. Exiting.", name, err)
```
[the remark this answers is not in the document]
emsixteeen: Hmm... Doesn't seem like it's possible to return an `error`:

[`SetWatchErrorHandler`](/kubernetes/client-go/blob/v0.27.4/tools/cache/shared_informer.go#L198) take a [`WatchErrorHandler`](/kubernetes/client-go/blob/v0.27.4/tools/cache/reflector.go#L133) – which does not return anything. If `SetWatchErrorHandler` fails, the code already returns an `error` for that.

With regard to [Replacing Fatal Calls](/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#replacing-fatal-calls), `klog.ErrorS` and `klog.FlushAndExit` is in later version of `klog`, but the operator is using [`v1`](/kubeflow/mpi-operator/blob/master/go.mod#L16), which doesn't have those methods ... 😕

emsixteeen:
> Users are not losing anything. Instead, they are getting more visibility into any possible failures, as they will see pods being recreated or restarted, causing them to investigate.

Oh, I see what you mean: in effect the pod is exiting and thus being forced into an unhealthy state ... Do you suggest then that the option be completely removed? This will cause the "current behavior" to change ...

pkg/controller/mpi_job_controller.go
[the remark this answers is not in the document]
emsixteeen: I made a mistake: `SetWatchErrorHandler` was calling `klog.Fatalf()` as well. Changed it to return an `error` if it fails.

Exploring some other ways a graceful shutdown can be initiated if the error handler goes get invoked (without calling `klog.Fatalf()`):

1. Upgrade to `klog/v2`
2. Call `kubeapiserver.RequestShutdown()`

emsixteeen:
> I don't expect anyone to be relying on the "current behavior".

I'm just trying not to break stuff 😄 

> So yes, I think we can just remove the option.

Removed

pkg/controller/mpi_job_controller.go
```diff
@@ -346,6 +346,34 @@ func NewMPIJobControllerWithClock(
...
+		"mpiJobInformer":        mpiJobInformer.Informer(),
+	}
+
+	for name, informer := range informers {
+		err := informer.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
+			// Pipe to default handler first, which just logs the error
+			cache.DefaultWatchErrorHandler(r, err)
+
+			if errors.IsUnauthorized(err) || errors.IsForbidden(err) {
+				klog.Fatalf("Unable to sync cache for informer %s: %s. Requesting controller to exit.", name, err)
```
alculquicondor, line 367: return error here too.

alculquicondor (commented):
Please squash

alculquicondor:
> I'm just trying not to break stuff 

Absolutely, I appreciate the amount of consideration for backwards-compatibility.

pkg/controller/mpi_job_controller.go
alculquicondor: oh true.
Still, please don't use `Fatalf`. Follow this guide https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#replacing-fatal-calls

emsixteeen: To follow this guide, `klog` needs to be upgraded to `klog/v2`. The methods `klog.ErrorS()` and `klog.FlushAndExit()` are not available in `klog/v1.0.0`.

Should `klog` be updated to `klog/v2` for this change?

alculquicondor: :(

Let's leave that for a follow up

/lgtm
/approve

Issue #621, opened 2024-02-05
Wrong host info in discover_hosts.sh


kuizhiqing:
When we create the mpijob with the same name in different namespaces, the host info may of one job may break in to the other one's discover_hosts.sh.

e.g. 
mpijob default/pi-1 and kubeflow/pi-1 may result in 
<img width="510" alt="image" src="https://github.com/kubeflow/mpi-operator/assets/2929527/55ef78ae-c69d-4c45-8d66-a6919c747c9c">

PR #622, opened 2024-02-05
Fix: add ns filter to podLister

kuizhiqing:
This PR fix https://github.com/kubeflow/mpi-operator/issues/621.

Since the podList have no namespace filter, it may capture the pods with same name but in different ns.


kuizhiqing:
merge https://github.com/kubeflow/mpi-operator/pull/623 first to solve CI issue.

alculquicondor (commented):
/lgtm
/approve

tenzen-y (commented):
/lgtm

PR #625, opened 2024-02-05
removing klog.Fatalf in favor of a shutdown request

emsixteeen:
Per [this discussion in #619](https://github.com/kubeflow/mpi-operator/pull/619#discussion_r1478715034), this is an option to change `klog.Fatalf()` to favor of something less harsh.

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -40,6 +40,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
+	kubeapiserver "k8s.io/apiserver/pkg/server"
 	batchinformers "k8s.io/client-go/informers/batch/v1"
 	coreinformers "k8s.io/client-go/informers/core/v1"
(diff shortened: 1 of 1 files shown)


alculquicondor:
I thought you had to upgrade klog?

emsixteeen:
This is **in lieu** of upgrading `klog`!

Being that the controller (in the `cmd` package) already uses [`kubeapiserver.SetupSignalHandler()`](https://github.com/kubeflow/mpi-operator/blob/master/cmd/mpi-operator/app/server.go#L99) to create the `stopCh`, we can also use [`kubeapiserver.RequestShutdown()`](https://github.com/kubernetes/apiserver/blob/v0.27.4/pkg/server/signal.go#L59) to close that channel.

In the [first conversation](https://github.com/kubeflow/mpi-operator/pull/619#discussion_r1478404176) I mentioned that in order not to call `klog.Fatalf()`, then either `klog` would need to be upgraded, or another option could be explored. This is the other option.

By the time I tested whether `kubeapiserver.RequestShutdown()` would work, the #619 was already merged! 😄

alculquicondor:
/close

let's leave it for now.

It might be better to wait to upgrade and convert all logs to structured logging.

emsixteeen:
> /close
> 
> let's leave it for now.
> 
> It might be better to wait to upgrade and convert all logs to structured logging.

👍

Issue #620, opened 2024-02-04
Running in a subset of namespaces


emsixteeen:
Currently it doesn't seem possible to run the `mpi-operator` in a **subset** of namespaces. It's either **all** namespaces or a **specific (single)** namespace.

This limitation looks like it comes from Kubernetes itself, where [generated] _Informers_ are either scoped to a single namespace using `<factory>.WithNamespace(...)`, or if no namespace is provider, defaults to `metav1.NamespaceAll`.

An example (real-world) use-case for running the `mpi-operator` in subset of namespaces is when it's deployed to a Kubernetes cluster with tightly controlled cluster-wide permissions. E.g. when obtaining cluster-wide access to `Secrets` is non-starter. In such a case, it's still possible to create a namespace-local `RoleBinding`, whereby access can be granted to the _Service Account_ running the `mpi-operator` for namespace-local secrets. However, because the `mpi-operator` operates only in either all namespaces or a single namespace, using namespace-local `RoleBindings` isn't expandable beyond a single namespace.

kuizhiqing:
It can be implemented with
```
mpiJobInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{FilterFunc: ...
```
in the enqeue process.

If we need it, I'd like to contribute.

emsixteeen:
It _**seems**_ that `cache.FilteringResourceEventHandler` will filter items once they have already been [successfully] listed / retrieved from the API server. 

The use-case / issue here is when the `mpi-operator` might not have permissions to list items in some namespace to begin with.

kuizhiqing:
As a feature of functionality, `FilteringResourceEventHandler` is the right way to **run controller in a subset of namespaces**, while the RBAC staffs is quite another thing.

emsixteeen:
@kuizhiqing totally agree that filtering is the way to go – assuming that RBAC permissions already allows for resources to be listed in all namespaces.

The current use-case / problem is that the Kubernetes cluster has a really restrictive RBAC configuration. With that being the case, because the controller is unable to list certain objects in the namespace to begin with (such as `Secrets`), getting to to filtering can never happen (classic chicken-and-egg situation: to filter we need to list, but to list, we need RBAC permissions, and RBAC is restricting those operations).

In trying to solve this issue, I came across https://github.com/maistra/xns-informer, which (using codegen) creates a multi-namespace informer out of existing informers.

[...]

PR #626, opened 2024-02-06
make namespace parsing and informers pluggable

emsixteeen:
This addresses #620 – in the case where the `mpi-operator` is running in a highly restrictive RBAC environment and the need to run `MPIJobs`  in only a subset of namespaces.

The changes provide:

1. A configurable namespace parser function, what will either use a default [and simple] parser, or a user-provided one.
2. Configurable _Informer_ functions that will created either the default `SharedInformerFactory` for each type of _Informer_ or user-provided ones.

### DIFF
--- a/cmd/mpi-operator/app/options/additional.go
+++ b/cmd/mpi-operator/app/options/additional.go
@@ -1,0 +1,38 @@
+package options
+
+import (
+	mpijobclientset "github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned"
+	informers "github.com/kubeflow/mpi-operator/pkg/client/informers/externalversions"
+	kubeinformers "k8s.io/client-go/informers"
+	kubeclientset "k8s.io/client-go/kubernetes"
+	schedclientset "sigs.k8s.io/scheduler-plugins/pkg/generated/clientset/versioned"
+	schedinformers "sigs.k8s.io/scheduler-plugins/pkg/generated/informers/externalversions"
+	volcanoclient "volcano.sh/apis/pkg/client/clientset/versioned"
+	volcanoinformers "volcano.sh/apis/pkg/client/informers/externalversions"
+)
+
+type NamespaceParserFunc func(namespace string, kubeClient kubeclientset.Interface) ([]string, error)
+
+type NamespaceOptions struct {
+	Namespaces NamespaceParserFunc
+}
+
+func DefaultNamespaceParser(namespace string, kubeClient kubeclientset.Interface) ([]string, error) {
+	return []string{namespace}, nil
+}
+
+type KubeInformerFunc func(namespaces []string, kubeClient kubeclientset.Interface) kubeinformers.SharedInformerFactory
+type MpiJobInformerFunc func(namespaces []string, mpiJobClient mpijobclientset.Interface) informers.SharedInformerFactory
+type VolcanoInformerFunc func(namespaces []string, volcanoClient volcanoclient.Interface) volcanoinformers.SharedInformerFactory
+type SchedulerPluginsInformerFunc func(namespaces []string, schedClient schedclientset.Interface) schedinformers.SharedInformerFactory
+
+type InformerOptions struct {
+	KubeInformer             KubeInformerFunc
+	MpiJobInformer           MpiJobInformerFunc
+	VolcanoInformer          VolcanoInformerFunc
+	SchedulerPluginsInformer SchedulerPluginsInformerFunc
+}
+type AdditionalOptions struct {
+	NamespaceOptions
+	InformerOptions
+}
--- a/cmd/mpi-operator/app/options/options.go
+++ b/cmd/mpi-operator/app/options/options.go
@@ -16,6 +16,7 @@ package options
 
 import (
 	"flag"
+	"github.com/kubeflow/mpi-operator/pkg/informers"
 	"os"
 
 	"github.com/kubeflow/mpi-operator/pkg/apis/kubeflow/v2beta1"
@@ -38,11 +39,21 @@ type ServerOption struct {
 	LockNamespace      string
 	QPS                int
 	Burst              int
+
+	NamespaceOptions
+	InformerOptions
 }
 
 // NewServerOption creates a new CMServer with a default config.
 func NewServerOption() *ServerOption {
 	s := ServerOption{}
+
+	s.Namespaces = DefaultNamespaceParser
+	s.KubeInformer = informers.DefaultKubeInformer
+	s.MpiJobInformer = informers.DefaultMpiJobInformer
+	s.VolcanoInformer = informers.DefaultVolcanoInformer
+	s.SchedulerPluginsInformer = informers.DefaultSchedulerPluginsInformer
+
 	return &s
 }
 
--- a/cmd/mpi-operator/app/server.go
+++ b/cmd/mpi-operator/app/server.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	kubeapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/server/healthz"
-	kubeinformers "k8s.io/client-go/informers"
 	kubeclientset "k8s.io/client-go/kubernetes"
 	clientgokubescheme "k8s.io/client-go/kubernetes/scheme"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -45,7 +44,6 @@ import (
 	"github.com/kubeflow/mpi-operator/cmd/mpi-operator/app/options"
 	mpijobclientset "github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned"
 	kubeflowscheme "github.com/kubeflow/mpi-operator/pkg/client/clientset/versioned/scheme"
-	informers "github.com/kubeflow/mpi-operator/pkg/client/informers/externalversions"
 	controllersv1 "github.com/kubeflow/mpi-operator/pkg/controller"
 	"github.com/kubeflow/mpi-operator/pkg/version"
 )
@@ -82,13 +80,6 @@ func Run(opt *options.ServerOption) error {
 		version.PrintVersionAndExit(apiVersion)
 	}
 
-	namespace := opt.Namespace
-	if namespace == corev1.NamespaceAll {
-		klog.Info("Using cluster scoped operator")
-	} else {
-		klog.Infof("Scoping operator to namespace %s", namespace)
-	}
-
 	// To help debugging, immediately log version.
 	klog.Infof("%+v", version.Info(apiVersion))
 
@@ -118,9 +109,23 @@ func Run(opt *options.ServerOption) error {
 	if err != nil {
 		return err
 	}
-	if !checkCRDExists(mpiJobClientSet, namespace) {
-		klog.Info("CRD doesn't exist. Exiting")
-		os.Exit(1)
+
+	namespaces, err := opt.Namespaces(opt.Namespace, kubeClient)
+	if err != nil {
+		return err
+	}
+
+	if namespaces[0] == corev1.NamespaceAll {
+		klog.Info("Using cluster scoped operator")
+	} else {
+		klog.Infof("Scoping operator to namespace %s", namespaces)
+	}
+
+	for _, namespace := range namespaces {
+		if !checkCRDExists(mpiJobClientSet, namespace) {
+			klog.Info("CRD doesn't exist. Exiting")
+			os.Exit(1)
+		}
 	}
 
 	// Add mpi-job-controller types to the default Kubernetes Scheme so Events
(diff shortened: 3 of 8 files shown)


tenzen-y (commented):
I was not sure why this change is needed since the current implementation (without this PR) also watchs only the dedicated namespace.

emsixteeen:
Here's the scenario ... 

* There are _N_ namespaces in the cluster (`ns1, ns2, ns3, ...`), and the cluster does not allow listing `Secrets` at the cluster level (which is enforced by some policy agent). 

* Per-namespace `RoleBindings` can be created to allow other _Subjects_ (such as `ServiceAccounts` in other namespaces) access to items like `Secrets` on a per-namespace level (i.e. `RoleBinding` vs `ClusterRoleBinding`).

* The desire is to run a single instance of the `mpi-operator` (in its own namespace), and have it watch for `MPIJobs` in _only_ a subset of the _N_ namespaces, e.g. `ns1` and `ns3`. In other words, the goal is to allow `MPIJobs` to run in some namespaces, but not all (more than one namespace, but not cluster-wide).

* Listing items like `Secrets` of the watched namespace(s) is a requirement of the `mpi-operator` (or it fails to run).

Because of the way _Informers_ work under the hood (_Informers_ implement the underlying watch functionality), you can only select **one** namespace to watch or **all** namespaces to watch. 

With the current implementation this scenario wouldn't seem to work ...

tenzen-y:
> Here's the scenario ...
> 
> * There are _N_ namespaces in the cluster (`ns1, ns2, ns3, ...`), and the cluster does not allow listing `Secrets` at the cluster level (which is enforced by some policy agent).
> * Per-namespace `RoleBindings` can be created to allow other _Subjects_ (such as `ServiceAccounts` in other namespaces) access to items like `Secrets` on a per-namespace level (i.e. `RoleBinding` vs `ClusterRoleBinding`).
> * The desire is to run a single instance of the `mpi-operator` (in its own namespace), and have it watch for `MPIJobs` in _only_ a subset of the _N_ namespaces, e.g. `ns1` and `ns3`. In other words, the goal is to allow `MPIJobs` to run in some namespaces, but not all (more than one namespace, but not cluster-wide).
> * Listing items like `Secrets` of the watched namespace(s) is a requirement of the `mpi-operator` (or it fails to run).
> 
> Because of the way _Informers_ work under the hood (_Informers_ implement the underlying watch functionality), you can only select **one** namespace to watch or **all** namespaces to watch.
> 
> With the current implementation this scenario wouldn't seem to work ...

@emsixteeen Uhm, I see. I understand your situation, but as I can see your code, the mpi-operator can only watch a single namespace the same as before, right?: https://github.com/kubeflow/mpi-operator/blob/a4c30dc76099bf3dbf04096fd7cc7f2fe359f7f1/pkg/informers/informers.go#L15-L19

emsixteeen:
> @emsixteeen Uhm, I see. I understand your situation, but as I can see your code, the mpi-operator can only watch a single namespace the same as before, right?:

That is correct – the PR does not change any current functionality or implementation. Rather, it provides an extension point to call different _InformerFactory_ functions for each of the _Informers_. 

It does this by replacing the actual call to `<SomePackage>.NewSharedInformerFactoryWithOptions(...)` function with a _function definition_. The function implementation makes the actual call to `<SomePackage>.NewSharedInformerFactoryWithOptions(...)`. 

The PR provides a _default_ function which behaves the same way as the current implementation, which is what you referenced.

The crux of the changes are where the function definition is defined:

https://github.com/kubeflow/mpi-operator/blob/a4c30dc76099bf3dbf04096fd7cc7f2fe359f7f1/cmd/mpi-operator/app/options/additional.go#L24-L34

Here is where / how it's used:

https://github.com/kubeflow/mpi-operator/blob/a4c30dc76099bf3dbf04096fd7cc7f2fe359f7f1/cmd/mpi-operator/app/server.go#L140-L141

For comparison, the current implementation:

https://github.com/kubeflow/mpi-operator/blob/a1ff84cbeb2c648e3e8f91f7c6a3bd7d2c768d8c/cmd/mpi-operator/app/server.go#L141-L142

___

Considering the scenario, the thought process and rationale for these changes:

**Background:** 

* Current _Informer_ packages uses Kubernetes codegen tools.
* Currently, the codegen tools create _Informers_ that watch either **one** or **all** namespaces.
* There's a package `xns-informer`[^1] which is built upon Kubernetes codegen tools that creates _Informers_ for **multiple** namespaces. Perhaps there are others.
* The _Informers_ created by the `xns-informer` package are experimental and are _usually_ – but _not guaranteed_ to be – API compatible[^2].

**Options:**

1. Determine that the scenario is not supportable. It's either one namespace or all namespaces.
2. Bring in the `xns-informer` package (or another package) as a dependency, and use its codegen tools to create _Informer_ packages.
3. Some middle ground: keep everything API compatible, while still providing a way to bring in a different _InformerFactory_.

**Thoughts:**

* The first option is the underlying issue being addressed!
* The second option seems a bit heavy-handed: bringing in a third-party experimental library seems counter-intuitive. 
* The third option seems like a better one: leave all functionality and implementation as is; everything remains API compatible; and extension points are added. If one wants to use experimental libraries, they can just extend the `mpi-operator` by hooking into it, while leaving everything else intact.

**Usability:**

In order to make use of the extension points, a **new** package is needed, which is composed using the `mpi-operator`[^3]. The upside to this is that one is free to swap _InformerFactory_ implementations. The downside is that these implementations live outside of the `mpi-operator` project and package. 

There's probably plenty more pros and cons which haven't even been touched upon ... Perhaps there's a some better way to go about this ...?

[^1]: https://github.com/maistra/xns-informer
[^2]: https://github.com/maistra/xns-informer?tab=readme-ov-file#cross-namespace-informers
[^3]: Example: https://github.com/emsixteeen/mpi-operator-xns

tenzen-y:
> > @emsixteeen Uhm, I see. I understand your situation, but as I can see your code, the mpi-operator can only watch a single namespace the same as before, right?:
> 
> That is correct – the PR does not change any current functionality or implementation. Rather, it provides an extension point to call different _InformerFactory_ functions for each of the _Informers_.
> 
> It does this by replacing the actual call to `<SomePackage>.NewSharedInformerFactoryWithOptions(...)` function with a _function definition_. The function implementation makes the actual call to `<SomePackage>.NewSharedInformerFactoryWithOptions(...)`.
> 
> The PR provides a _default_ function which behaves the same way as the current implementation, which is what you referenced.
> 
> The crux of the changes are where the function definition is defined:
> 
> https://github.com/kubeflow/mpi-operator/blob/a4c30dc76099bf3dbf04096fd7cc7f2fe359f7f1/cmd/mpi-operator/app/options/additional.go#L24-L34
> 
> Here is where / how it's used:
> 
> https://github.com/kubeflow/mpi-operator/blob/a4c30dc76099bf3dbf04096fd7cc7f2fe359f7f1/cmd/mpi-operator/app/server.go#L140-L141
> 
> For comparison, the current implementation:
> 
> https://github.com/kubeflow/mpi-operator/blob/a1ff84cbeb2c648e3e8f91f7c6a3bd7d2c768d8c/cmd/mpi-operator/app/server.go#L141-L142
> 
> Considering the scenario, the thought process and rationale for these changes:
> 
> **Background:**
> 
> * Current _Informer_ packages uses Kubernetes codegen tools.
> * Currently, the codegen tools create _Informers_ that watch either **one** or **all** namespaces.
> * There's a package `xns-informer`[1](#user-content-fn-1-acc8c0a34272b4ca63548980bccace90) which is built upon Kubernetes codegen tools that creates _Informers_ for **multiple** namespaces. Perhaps there are others.
> * The _Informers_ created by the `xns-informer` package are experimental and are _usually_ – but _not guaranteed_ to be – API compatible[2](#user-content-fn-2-acc8c0a34272b4ca63548980bccace90).
> 
> **Options:**
> 
> 1. Determine that the scenario is not supportable. It's either one namespace or all namespaces.
> 2. Bring in the `xns-informer` package (or another package) as a dependency, and use its codegen tools to create _Informer_ packages.
> 3. Some middle ground: keep everything API compatible, while still providing a way to bring in a different _InformerFactory_.
> 
> **Thoughts:**
> 
> * The first option is the underlying issue being addressed!
> * The second option seems a bit heavy-handed: bringing in a third-party experimental library seems counter-intuitive.
> * The third option seems like a better one: leave all functionality and implementation as is; everything remains API compatible; and extension points are added. If one wants to use experimental libraries, they can just extend the `mpi-operator` by hooking into it, while leaving everything else intact.
> 
> **Usability:**
> 
> In order to make use of the extension points, a **new** package is needed, which is composed using the `mpi-operator`[3](#user-content-fn-3-acc8c0a34272b4ca63548980bccace90). The upside to this is that one is free to swap _InformerFactory_ implementations. The downside is that these implementations live outside of the `mpi-operator` project and package.
> 
> There's probably plenty more pros and cons which haven't even been touched upon ... Perhaps there's a some better way to go about this ...?
> 
> ## Footnotes
> 1. https://github.com/maistra/xns-informer [↩](#user-content-fnref-1-acc8c0a34272b4ca63548980bccace90)
> 2. https://github.com/maistra/xns-informer?tab=readme-ov-file#cross-namespace-informers [↩](#user-content-fnref-2-acc8c0a34272b4ca63548980bccace90)
> 3. Example: https://github.com/emsixteeen/mpi-operator-xns [↩](#user-content-fnref-3-acc8c0a34272b4ca63548980bccace90)

Thank you for this clarifications.
This means that you want to implement a separate mpi-operator with x-ns informers in external repositories like https://github.com/emsixteeen/mpi-operator-xns.
So, you want to export informer factories in this mpi-operator repository, right?

TBH, I don't like such extension points for external mpi-operator.
I believe that we should make an effort in this upstream mpi-operator.

Also, if it's challenging to make efforts for you in this repository or upstream kubernetes, you can fork this repository, and you can directly improve the forked mpi-operator with x-ns informers.

@alculquicondor WDYT?

emsixteeen:
> But, I'm not sure that the controller-runtime will really resolve this issue.

I don't think it will resolve it ... I haven't checked the `controller-runtime` code too deeply, but I'm assuming that it uses that standard `client-go` (i.e. generated) _Informers_. 

If it's the case – that it's using the standard generated _Informers_ – then that's the key issue: an _Informer_ will List either **one** namespace or **all** namespaces. The filtering is done post-facto – i.e. the _Informer_ will get objects from **all** namespaces, but the filter will only act upon interested namespace(s).

The catch there is that if RBAC rules don't allow for some item to be listed across **all** namespaces, the _Informer_ will keep getting a permission error, and the `HasSycned` calls won't return `true`.

emsixteeen:
> Thank you for this clarifications. This means that you want to implement a separate mpi-operator with x-ns informers in external repositories like https://github.com/emsixteeen/mpi-operator-xns. So, you want to export informer factories in this mpi-operator repository, right?

It's not that I'm looking to implement another operator, inasmuch as be afforded functionality that is not available in the current design. (Noting that the current design is more an artifact of how Kubernetes codegen works, and **not** directly related to the design of the `mpi-operator` itself.)

>
> TBH, I don't like such extension points for external mpi-operator. I believe that we should make an effort in this upstream mpi-operator.

Totally understand... The main reason this PR creates extension points was so that there would be no need to bring in (and create dependencies upon) experimental packages (such as `xns-informer`, or others). 

On one hand, creating extension points can become a liability in the long term (needing to support it, side effects, risking breaking other things down the road, etc.) ... On the other hand, I don't know of another way to do this without creating additional dependencies and also obtaining this functionality (barring upstream code changes in how `client-go` and Kuberentes codegen works). Having these extension points though, creates no change in functionality in the upstream operator. Potential issues / side effects _can_ be introduced if the extension points were used – but that would be the responsibility of the extender. In other words, even with these extension points in place, nothing changes in the upstream operator.

Regardless, I'm completely open to bringing this functionality directly into [this] upstream `mpi-operator`... What might be some next-best steps, or ideas, to implementing this? Is it to OK using to use `Informers` that are codegen'd by something other than Kubernetes codgen? (Again, the initial focus was to _not_ bring in additional dependencies that are out of the core Kubernetes ecosystem.)

Here's an early prototype of what bringing in `xns-informer` generated _Informers_ might look like: https://github.com/emsixteeen/mpi-operator/tree/xns-informers

>
> Also, if it's challenging to make efforts for you in this repository or upstream kubernetes, you can fork this repository, and you can directly improve the forked mpi-operator with x-ns informers.

Given the value the project has delivered, I'd really rather the effort be applied to the upstream project... 

> 
> @alculquicondor WDYT?

tenzen-y:
@emsixteeen I meant that I disagree with both exporting Informer for external projects and using the xns-informer in this project.
If you want to use the features that have xns-informer, I think you should discuss to add the features to kubernetes core features in [the upstream Kubernetes](https://github.com/kubernetes).

If you don't want to such a discussion in the kubernetes core project, you can just fork this repository and then you can extend the mpi-operator using the `xns-informer`.

I think importing not maintainable library (xns-informer) should avoid as much as possible, and we should not increase maintenance costs by exporting functions for the external projects.

PR #627, opened 2024-02-27
Deprecated pointer, use ptr instead

kuizhiqing:
Since `https://pkg.go.dev/k8s.io/utils/pointer` is deprecated, we use `ptr` to replace `pointer`.


tenzen-y (commented):
/lgtm
/assign @alculquicondor

PR #628, opened 2024-02-29
Fix: no overwrite when run launcher as worker

kuizhiqing:
Default implementation overwrite envs `NVIDIA_VISIBLE_DEVICES` and `NVIDIA_DRIVER_CAPABILITIES` which is not work in the `RunLauncherAsWorker` case with GPU.


tenzen-y (commented):
Implementation is lgtm
Could you extend unit tests here?: https://github.com/kubeflow/mpi-operator/blob/f92b9c7e74cb8618c8c153ccb3921bd0b160782c/pkg/controller/mpi_job_controller_test.go#L1195

tenzen-y (commented):
Thanks!

/lgtm
/assign @alculquicondor

alculquicondor (commented):
/approve

PR #630, opened 2024-03-26
Remove unnecessary RBAC rule for mpijobs-admin***

vishvajit79:
***Update cluster-role.yaml to remove the unnecessary RBAC rule for mpijobs-admin.***


manifests/base/cluster-role.yaml
```diff
@@ -126,7 +126,6 @@ metadata:
   name: kubeflow-mpijobs-edit
   labels:
     rbac.authorization.kubeflow.org/aggregate-to-kubeflow-edit: "true"
-    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-mpijobs-admin: "true"
```
alculquicondor, line 129: why remove this? An admin might need to perform edits in case of emergencies.

vishvajit79: If that's the case, then we should define rules for admin. Right now, rules for admin are empty array and when the admin annotation is added to edit clusterrole, it keeps overriding admin role. If we expect the admin to have slightly different role than edit then it would make sense to keep both otherwise I would suggest removing the admin role

alculquicondor: @terrytangyuan @rongou which practices are you following in the training-operator?
I don't have all the context to make a decision here.

tenzen-y: @alculquicondor IIRC, The training-operator doesn't use RBAC aggregations: https://github.com/kubeflow/training-operator/blob/2eff94ea8131879c7175ba6ae9e3d7d098f92f85/manifests/base/rbac/role.yaml

I guess that this setting seems to be only for the MPIJob v2.

tenzen-y: It makes sense. Actually, I also don't have any context...

alculquicondor: Maybe @johnugeorge has some context of the original intent?

terrytangyuan: I don't have context either. I wonder if it's for integration with UI or some other front-ends.

Issue #646, opened 2024-08-05
Add support for the managedBy field


mimowo:
### What you would like to be added?

The support for the `managedBy` field which can delegate reconciliation from built-in controller, to a custom one.

The semantics of the field are:
- whenever the value is set, and it does not point to the built-in operator, then skip reconciliation
- the field is immutable
- the field is not defaulted

### Why is this needed?

For context, we have in Kueue the effort (see [Support Kubeflow Jobs in MultiKueue](https://github.com/kubernetes-sigs/kueue/issues/2552)) to support kubeflow-training (it will include MPIJob), but it will not be complete without the support for managedBy.

The complete support for the users of [MultiKueue](https://kueue.sigs.k8s.io/docs/tasks/manage/setup_multikueue/) (multi-cluster Kueue) means:
- simpler installation (just follow the standard installation path), otherwise only installation of MPIJob CRDs is required
- support for mixed setup in one cluster - some MPIJob could be run by MultiKueue and some by the default operator

The efforts to support the field in:
- [k8s Job](https://github.com/kubernetes/enhancements/issues/4368) (alpha)
- [JobSet](https://github.com/kubernetes-sigs/jobset/issues/440) (complete)

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

mimowo:
/cc @andreyvelich @tenzen-y 
Would you like to include this feature?

mimowo:
xref for the similar issue in the training operator: https://github.com/kubeflow/training-operator/issues/2193

tenzen-y:
As we discussed in kubeflow community meeting, basically, I'm with this feature.
But, I would like to know what @alculquicondor thinks since maybe I don't have enough bandwidth to review this feature.

alculquicondor:
+1 to this feature.

I also can't promise review bandwidth, but it's something that needs to be done for consistency with the other APIs.

tenzen-y:
@mszadkow Do you have any plans to work on this issue?

PR #650, opened 2024-09-21
Introduce ManagedBy field in RunPolicy

mszadkow:
Add support for ManagedBy to be able delegate reconciliation process to external controller.

Fixes https://github.com/kubeflow/mpi-operator/issues/646


pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -131,6 +139,16 @@ type RunPolicy struct {
 	// Defaults to false.
 	// +kubebuilder:default:=false
 	Suspend *bool `json:"suspend,omitempty"`
+
+	// ManagedBy is used to indicate the controller or entity that manages a job.
```
tenzen-y, line 143:
```suggestion
	// ManagedBy is used to indicate the controller or entity that manages a MPIJob.
```

Let's identify which Job is delegated to the external controller.
Because the mpi-operator has batch/v1 Job concept.

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -131,6 +139,16 @@ type RunPolicy struct {
...
+	// The value must be either an empty, 'kubeflow.org/mpi-operator' or
+	// 'kueue.x-k8s.io/multikueue'.
+	// The mpi-operator reconciles a job which doesn't have this
```
tenzen-y, line 146:
```suggestion
	// The mpi-operator reconciles a MPIJob which doesn't have this
```

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -131,6 +139,16 @@ type RunPolicy struct {
...
+	// field at all or the field value is the reserved string
+	// 'kubeflow.org/mpi-operator', but delegates reconciling the job
```
tenzen-y, line 148:
```suggestion
	// 'kubeflow.org/mpi-operator', but delegates reconciling the MPIJob
```

pkg/controller/mpi_job_controller_test.go
```diff
@@ -504,6 +504,18 @@ func TestDoNothingWithInvalidMPIJob(t *testing.T) {
...
+	f := newFixture(t, "")
+	var replicas int32 = 1
+	startTime := metav1.Now()
+	completionTime := metav1.Now()
+	mpiJob := newMPIJob("test", &replicas, &startTime, &completionTime)
+	mpiJob.Spec.MPIImplementation = kubeflow.MPIImplementationOpenMPI
+	mpiJob.Spec.RunPolicy.ManagedBy = ptr.To(kubeflow.MultiKueueController)
+	f.setUpMPIJob(mpiJob)
+	f.run(getKey(mpiJob, t))
+}
```
tenzen-y, line 517: Could we verify if secret, launcher, and workers here?

test/e2e/mpi_job_test.go
```diff
@@ -164,6 +165,27 @@ var _ = ginkgo.Describe("MPIJob", func() {
 				mpiJob := createJobAndWaitForCompletion(mpiJob)
 				expectConditionToBeTrue(mpiJob, kubeflow.JobSucceeded)
 			})
+
+			ginkgo.It("should not be updated when managed externaly, only created", func() {
+				mpiJob.Spec.RunPolicy.ManagedBy = ptr.To(kubeflow.MultiKueueController)
+				ctx := context.Background()
+				mpiJob = createJob(ctx, mpiJob)
+
+				gomega.Consistently(func() error {
```
tenzen-y, line 174:
```suggestion
				gomega.Eventually(func() error {
```

I guess that the `Eventually` should be enough since this verification is just validation testing, right?

test/e2e/mpi_job_test.go
```diff
@@ -164,6 +165,27 @@ var _ = ginkgo.Describe("MPIJob", func() {
...
+					mpiJob = updatedJob
+					return nil
+				}, 5*time.Second, waitInterval).Should(gomega.Succeed())
+
+				// job should be created, but status should not be updated neither for create nor for any other status
+				condition := getJobCondition(mpiJob, kubeflow.JobCreated)
+				gomega.Expect(condition).To(gomega.BeNil())
+				condition = getJobCondition(mpiJob, kubeflow.JobSucceeded)
+				gomega.Expect(condition).To(gomega.BeNil())
+			})
```
tenzen-y, line 188: Could you verify if Secret, Launcher, and Workers have never been created?

pkg/controller/mpi_job_controller_test.go
```diff
@@ -504,6 +504,18 @@ func TestDoNothingWithInvalidMPIJob(t *testing.T) {
...
+
```
tenzen-y, line 518: Additionally, could you add if the MPIJobs with `kubeflow.org/mpi-operator` are reconciled by the mpi-operator?

test/e2e/mpi_job_test.go
mszadkow: eventually assumes something will change over time and we wait for the condition, consistently proves us nothing changed - we want the latter

pkg/controller/mpi_job_controller_test.go
mszadkow: I could add the mpi-operator in `newMPIJobCommon`, so all other tests verify that, although they should already as this is default value
```
	if policy.ManagedBy == nil {
		policy.ManagedBy = ptr.To(KubeflowJobController)
	}
```

mszadkow: I figured that proving no kubeActions should be sufficient, wdyt?

test/e2e/mpi_job_test.go
mszadkow: sure, done

pkg/controller/mpi_job_controller_test.go
mszadkow: done

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -131,6 +139,16 @@ type RunPolicy struct {
 	// Defaults to false.
 	// +kubebuilder:default:=false
 	Suspend *bool `json:"suspend,omitempty"`
+
+	// ManagedBy is used to indicate the controller or entity that manages a MPIJob.
+	// The value must be either an empty, 'kubeflow.org/mpi-operator' or
+	// 'kueue.x-k8s.io/multikueue'.
```
alculquicondor, line 145: This API cannot say that these are the only possible values. This API shouldn't even be aware that Kueue exists.

The only thing that it could say is that if the value is not `kubeflow.org/mpi-operator` then it's not managed by this controller. See https://github.com/kubernetes/api/blob/71385f038c1097af36f3d2f68b415860b866c1f8/batch/v1/types.go#L473 for an example of how it should look like.

pkg/apis/kubeflow/validation/validation.go
```diff
@@ -39,8 +39,11 @@ var (
 
 	validRestartPolicies = sets.NewString(
 		string(kubeflow.RestartPolicyNever),
-		string(kubeflow.RestartPolicyOnFailure),
-	)
+		string(kubeflow.RestartPolicyOnFailure))
+
+	validManagedBy = sets.NewString(
```
alculquicondor, line 44: The validation should be similar to https://github.com/kubernetes/kubernetes/blob/b4ce3a55bc7218a6710ff9340dccd403eee5e293/pkg/apis/batch/validation/validation.go#L220

pkg/controller/mpi_job_controller_test.go
```diff
@@ -504,6 +509,21 @@ func TestDoNothingWithInvalidMPIJob(t *testing.T) {
 	f.run(getKey(mpiJob, t))
 }
 
+func TestDoNothingWithMPIJobManagedExternally(t *testing.T) {
```
alculquicondor, line 512: Also validate that a nil value the ManagedBy field leads to processing.

This is important for jobs that were created before the ManagedBy field existed

test/e2e/mpi_job_test.go
```diff
@@ -164,8 +165,43 @@ var _ = ginkgo.Describe("MPIJob", func() {
...
+				mpiJob = createJob(ctx, mpiJob)
+
+				gomega.Consistently(func() error {
+					updatedJob, err := mpiClient.KubeflowV2beta1().MPIJobs(mpiJob.Namespace).Get(ctx, mpiJob.Name, metav1.GetOptions{})
+					if err != nil {
+						return err
+					}
+					mpiJob = updatedJob
+					return nil
+				}, 5*time.Second, waitInterval).Should(gomega.Succeed())
```
alculquicondor, line 181: this is too much time. Maybe just 2 seconds.

pkg/apis/kubeflow/v2beta1/types.go
tenzen-y: Here is background of the reason why we limit the acceptable values in the managedBy resource.
Ideally, we should enhance the status validations similar to batch/Job and allow managedBy field to be specified arbitrary values
so that we can provide the appropriate rail.

But, the proposal was rejected in the training-operator since the TrainigOperator community is working on the v2 API, and they have no enough bandwidth to implement the additional status validation implementation. And, the v2 API will provided the non restricted managedBy feature same as batch/Job.

So, I think we can take either approach: 1. enhance status validation and allow managedBy to specify arbitrary values 2. Limit the acceptable values kueue or mpi-operator.

If anyone can lead the design and handle the status validation enhancement, I'm ok with opt 1.

alculquicondor: I see, that's unfortunate.

Let's just do the same as the training-operator.

For future reference, the status validations refer to the equivalent of this https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/4368-support-managed-by-for-batch-jobs#job-status-validation

tenzen-y: Sorry, I forgot to add the contextual issue 😓 
https://github.com/kubeflow/training-operator/issues/2193

tenzen-y: > I see, that's unfortunate.
> 
> Let's just do the same as the training-operator.
> 
> For future reference, the status validations refer to the equivalent of this https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/4368-support-managed-by-for-batch-jobs#job-status-validation

Yes, ideally, we should do it. Maybe, we can leave the dream on the TrainJob 😓

pkg/controller/mpi_job_controller_test.go
mszadkow: Sure, we could add extra test case, but all other tests use mpiJob returned from `newMPIJob`, where managedBy is nil by default.
Although I could enhance that by adding it explicitly:
```go
func newMPIJobCommon(name string, startTime, completionTime *metav1.Time) *kubeflow.MPIJob {
...
	mpiJob := &kubeflow.MPIJob{
...
		Spec: kubeflow.MPIJobSpec{
			RunPolicy: kubeflow.RunPolicy{
				CleanPodPolicy: &cleanPodPolicyAll,
				ManagedBy: nil,
			},
```
wdyt?

test/e2e/mpi_job_test.go
mszadkow: Yes, sure, it should be sufficient

pkg/apis/kubeflow/validation/validation.go
alculquicondor: There is no point on following my suggestion if the values are fixed. Forget about this for now.

pkg/controller/mpi_job_controller_test.go
alculquicondor: No, that's not necessary. Thanks for checking.

alculquicondor:
@tenzen-y anything else to add?

test/e2e/mpi_job_test.go
tenzen-y: That makes sense. Thanks!

test/e2e/mpi_job_test.go
```diff
@@ -164,8 +165,43 @@ var _ = ginkgo.Describe("MPIJob", func() {
...
+					mpiJob = updatedJob
+					return nil
+				}, 2*time.Second, waitInterval).Should(gomega.Succeed())
+
+				// job should be created, but status should not be updated neither for create nor for any other status
+				condition := getJobCondition(mpiJob, kubeflow.JobCreated)
+				gomega.Expect(condition).To(gomega.BeNil())
+				condition = getJobCondition(mpiJob, kubeflow.JobSucceeded)
+				gomega.Expect(condition).To(gomega.BeNil())
+				launcherPods, err := getLauncherPods(ctx, mpiJob)
```
tenzen-y, line 188: Could you verify if the Launcher Job doesn't exist instead of the Launcher Pod?

test/integration/mpi_job_controller_test.go
```diff
@@ -822,6 +823,96 @@ func TestMPIJobWithVolcanoScheduler(t *testing.T) {
...
+		}
+		// 3. There should be no conditions, even the one for create
+		if mpiJobHasCondition(mpiJob, kubeflow.JobCreated) {
+			t.Errorf("MPIJob shouldn't have any condition")
+		}
+		// 4. No Pods or Services created
+		pods, err := getPodsForJob(ctx, s.kClient, mpiJob)
+		if err != nil {
+			t.Fatalf("Failed getting pods for the job: %v", err)
+		}
```
tenzen-y, line 898: Could we verify if the Launcher Job doesn't exist as well since the MPIOperator will create the Launcher batch/v1 Job and Worker Pods?

tenzen-y (commented):
Thanks. I left minor comments.
Otherwise lgtm

test/e2e/mpi_job_test.go
mszadkow: done

test/integration/mpi_job_controller_test.go
mszadkow: done

pkg/apis/kubeflow/v2beta1/default.go
```diff
@@ -53,6 +53,9 @@ func setDefaultsRunPolicy(policy *RunPolicy) {
 	if policy.CleanPodPolicy == nil {
 		policy.CleanPodPolicy = ptr.To(CleanPodPolicyNone)
 	}
+	if policy.ManagedBy == nil {
```
mimowo, line 56: Why are we defaulting the field? It is a decision in other Job CRDs not to default it, so I would suggest to keep it this way, for consistency. It can also be helpful for MultiKueue. If we default the field here, then the MultiKueue webhook will need to override the value, rather than just set.

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -131,6 +139,16 @@ type RunPolicy struct {
...
+	// The mpi-operator reconciles a MPIJob which doesn't have this
+	// field at all or the field value is the reserved string
+	// 'kubeflow.org/mpi-operator', but delegates reconciling the MPIJob
+	// with 'kueue.x-k8s.io/multikueue' to the Kueue.
+	// The field is immutable.
```
mimowo, line 150: Should add `// +optional` tag

test/integration/mpi_job_controller_test.go
```diff
@@ -822,6 +823,96 @@ func TestMPIJobWithVolcanoScheduler(t *testing.T) {
...
+	mpiJob, err = s.mpiClient.KubeflowV2beta1().MPIJobs(s.namespace).Create(ctx, mpiJob, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("Failed sending job to apiserver: %v", err)
+	}
+	backoff := wait.Backoff{
+		Duration: moderateTimeout,
+		Factor:   1.0,
+		Steps:    3,
+	}
+	err = wait.ExponentialBackoff(backoff, func() (bool, error) {
```
mimowo, line 884: What is the purpose of using `ExponentialBackoff` here? Can we just wait 1s before doing the asserts? I assume this will be enough to fail if ManagedBy wasn't suppored.

test/e2e/mpi_job_test.go
```diff
@@ -164,8 +166,46 @@ var _ = ginkgo.Describe("MPIJob", func() {
...
 				expectConditionToBeTrue(mpiJob, kubeflow.JobSucceeded)
 			})
-		})
 
+			ginkgo.It("should not be updated when managed externaly, only created", func() {
+				mpiJob.Spec.RunPolicy.ManagedBy = ptr.To(kubeflow.MultiKueueController)
+				ctx := context.Background()
+				mpiJob = createJob(ctx, mpiJob)
+
+				gomega.Consistently(func() error {
```
mimowo, line 175: I don't think we need to retry the Get with Consistently. I think waiting 1s before the Get should be enough. If ManagedBy wasn't respected then the Get would return updated object.

test/integration/mpi_job_controller_test.go
mszadkow: It's not very common to use plain sleep in tests, but sure it does the job

mimowo: By using `ExponentialBackoff` you do the sleep anyway, just hidden behind the scenes, and the reason is not clear. 

I think it is better to be transparent and comment on why we sleep, or name the var in a self-explanatory way. 

This is the approach we take in the core k8s: https://github.com/kubernetes/kubernetes/blob/0b91def6bcbdde82206ec779791762fb6bc337af/test/integration/job/job_test.go#L2108.

test/integration/mpi_job_controller_test.go
```diff
@@ -876,41 +876,33 @@ func TestMPIJobManagedExternally(t *testing.T) {
...
-		}
-		if svcs != nil {
-			t.Fatalf("There should be no services from job: %v", svcs)
-		}
-		return false, nil
-	})
-	if !wait.Interrupted(err) {
-		t.Fatalf("Failed to verify externally managed mpiJob: %v", err)
+
+	time.Sleep(moderateTimeout)
```
mimowo, line 880: This is not really used as timeout here. Instead of using `moderateTimeout `use a variable which reflects the intention, sleepDurationForControllerLatency as in https://github.com/kubernetes/kubernetes/blob/master/test/integration/job/job_test.go#L2108 sgtm.

tenzen-y (commented):
Thanks!
/approve

Leave lgtm on @alculquicondor

pkg/apis/kubeflow/v2beta1/types.go
```diff
@@ -131,6 +139,17 @@ type RunPolicy struct {
 	// Defaults to false.
 	// +kubebuilder:default:=false
 	Suspend *bool `json:"suspend,omitempty"`
+
+	// ManagedBy is used to indicate the controller or entity that manages a MPIJob.
+	// The value must be either an empty, 'kubeflow.org/mpi-operator' or
```
alculquicondor, line 144:
```suggestion
	// The value must be either empty, 'kubeflow.org/mpi-operator' or
```

alculquicondor (commented):
/lgtm

pkg/apis/kubeflow/v2beta1/types.go
tenzen-y: @mszadkow Could you address this comment in a separate PR?

mszadkow: Sure, will be addressed asap

PR #653, opened 2024-10-10
Upgrade the scheduler-plugins to v0.29.8

tenzen-y:
I upgraded the scheduler-plugins version to the latest one.


alculquicondor:
/lgtm
/approve

Could scheduler-plugins block us from upgrading k8s libraries further?

tenzen-y:
> /lgtm /approve
> 
> Could scheduler-plugins block us from upgrading k8s libraries further?

I guess that the scheduler-plugins do not block us since we use only the scheduler-plugins API.

Issue #654, opened 2024-10-10
Release v0.6.0 requirements


tenzen-y:
I would propose a new v0.6.0 release once we resolve the following tasks.

### Dependency Update

- [x] scheduler-plugins: https://github.com/kubeflow/mpi-operator/pull/653
- [x] volcano: #659 @tenzen-y 
- [x] K8s
    - [x] 1.30: https://github.com/kubeflow/mpi-operator/pull/657 @tenzen-y 
    - [x] 1.31: https://github.com/kubeflow/mpi-operator/pull/664 @ArangoGutierrez 
- [x] Debian: #661 @tenzen-y 

Note that we will update the K8s libraries step by step so that users can use the dedicated K8s version once they face issues depending on the library version.

### Feature

- [x] ManagedBy: https://github.com/kubeflow/mpi-operator/pull/650 @mszadkow

tenzen-y:
cc @alculquicondor @terrytangyuan @mimowo @klueska

terrytangyuan:
Great. Thanks!

mimowo:
Awesome, thanks!

ArangoGutierrez:
Lgtm

tenzen-y:
The v0.6.0 has already been released. Thanks all!
/close

PR #655, opened 2024-10-10
Bump K8s to 1.31

ArangoGutierrez:
This PR also adds a new example on how to use DRA to claim GPU's for an MPI job


tenzen-y:
@ArangoGutierrez As I mentioned in https://github.com/kubeflow/mpi-operator/issues/654, could you upgrade the k8s version step by step?

examples/v2beta1/dra/README.md
```diff
@@ -0,0 +1,5 @@
+# DRA enabled Kubernetes
```
kannon92, line 1: I think this should be a separate PR once the upgrade goes in.

DRA support on OSS projects is quite difficult at the moment. You need to guarantee that someone has the feature gates on (which is not yet possible in any k8s cluster by default), nvidia dra driver is installed and hope that the APIs for this don't change over time.

Also, these examples would require nvidia gpus of course. I tend to prefer examples on OSS projects to be CPU based in cases where performance doesn't matter. I don't know how this project keeps examples up to date and working. In other areas, we are exploring the idea of using the sample driver for DRA to at least show how DRA works without needing GPU access to actually run this.

tenzen-y: +1
Additionally, I do not want to block the new release by this PR.
So, could you focus on lib bumping in the PR?

ArangoGutierrez: Dropped

Makefile
tenzen-y, line 1: What is the reason to modify this Makefile?
I guess that without changes, it should work fine.

go.mod
```diff
@@ -1,114 +1,115 @@
 module github.com/kubeflow/mpi-operator
 
-go 1.22.0
+go 1.23.0
+
+toolchain go1.23.1
```
tenzen-y, line 5:
```suggestion
go 1.23
```

AFAIK, since Go 1.23, we do not need to specify the toolchain version.

pkg/apis/kubeflow/v2beta1/zz_generated.deepcopy.go
tenzen-y, line 1: Could you keep this?

manifests/base/kubeflow.org_mpijobs.yaml
tenzen-y, line 1: Could you keep this?

hack/update-codegen.sh
tenzen-y, line 1: What is the reason for changing this script?

PR #573, opened 2023-06-22
Update version of debian for Docker images

sheevy:
Bullseye -> Bookworm

PR #661, opened 2024-10-12
Introduce debian bookworm

tenzen-y:
I upgraded the Debian bullseye to the bookworm. The bookworm allows us to use the following MPI versions.

- OpenMPI: [v4.1.0](https://packages.debian.org/bullseye/openmpi-bin) ->  [v4.1.4](https://packages.debian.org/bookworm/openmpi-bin)
- IntelMPI: 2021.13.1 -> 2021.13.1 (without changes)
- MPICH: [3.4.1](https://packages.debian.org/bullseye/mpich) -> [4.0.2](https://packages.debian.org/bookworm/mpich)

### Side Effect

The new bookworm contains the v4 MPICH, but in the MPICH v4, it seems that we need to obviously verify if all Ranks are finalized using `MPI_Barrier()` before `MPI_Finalize()`. If we do not verify it, the rendezvous point will be closed before all Ranks are finalized, and then the Worker Ranks will be stuck forever, even if all Ranks succeeded, as observed at #573.

There is a similar situation: https://github.com/pmodels/mpich/issues/6584

Fixes: #573

### DIFF
--- a/build/base/Dockerfile
+++ b/build/base/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:bullseye
+FROM debian:bookworm
 
 ARG port=2222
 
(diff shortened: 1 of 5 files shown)


ArangoGutierrez (approved):
lgtm

PR #666, opened 2024-10-15
Fix the 'printf: non-constant format string in call to fmt.Errorf (govet)' lint errors

tenzen-y:
I fixed lint errors as we can see in https://github.com/kubeflow/mpi-operator/actions/runs/11353824606/job/31579702079, and I removed the `--new-from-rev=origin/master` option from golangci-lint so that we can avoid the master branch error.

```
pkg/controller/mpi_job_controller.go:886:26: printf: non-constant format string in call to fmt.Errorf (govet)
                return nil, fmt.Errorf(msg)
                                       ^
```

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -20,4 +20,3 @@ jobs:
         with:
           version: v1.61.0
           args: --timeout=5m
-          only-new-issues: true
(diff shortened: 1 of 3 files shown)


terrytangyuan (approved):
This is great!

/lgtm
/approve

PR #674, opened 2025-01-01
Expose job controller's workqueue rate limiting configs

roteme-runai:
### Issue:
The `mpi-operator` does not handle a large number of MPI jobs effectively. Specifically, when creating a significant number of jobs (e.g., 100), the operator experiences delays in job management, leading to the following issues:
- Excessive time taken to create all the pods required for the jobs.
- Significant delay between a job's completion and the corresponding status update.
- Delayed cleanup of pods (per `CleanPodPolicy`) due to the lag in status updates.

### Root Cause:
The controller uses a workqueue with a default rate limiter configuration that is not adjustable via operator options. This is in contrast to other load-related configurations (e.g., `threadiness`, `qps`, etc.), which are user-configurable. The low default rate-limiting settings result in insufficient parallel processing, thereby delaying job handling.

### Proposed Solution:
To address this, I propose exposing the controller's rate-limiting settings as user-configurable options. This change would allow users to adjust the rate limiter based on their specific usage requirements, expected scale, and system capabilities.
The solution has been tested and verified in a production environment and has demonstrated improved handling of bigger-scale MPI jobs.

### Backporting Request:
If this fix is approved, I kindly request its inclusion in a new release, ideally in versions from v0.6 onward.

Thank you for your time!


cmd/mpi-operator/app/options/options.go
```diff
@@ -75,4 +77,7 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 
 	fs.IntVar(&s.QPS, "kube-api-qps", 5, "QPS indicates the maximum QPS to the master from this client.")
 	fs.IntVar(&s.Burst, "kube-api-burst", 10, "Maximum burst for throttle.")
+
+	fs.IntVar(&s.ControllerRateLimit, "controller-queue-rate-limit", 10, "Rate limit of the controller events queue .")
+	fs.IntVar(&s.ControllerBurst, "controller-queue--burst", 100, "Maximum burst of the controller events queue.")
```
mlsorensen, line 82: Is the double hyphen a typo?

roteme-runai: it was. fixed

mlsorensen (approved):
LGTM. Confirmed the defaults for the new flags match the previous default limit and burst, shouldn't change behavior.

mchmarny:
@tenzen-y @jlewi @rongou can you help review this?

jlewi:
I'm not involved anymore so I'm probably not the right person to be reviewing this.

rongou:
/lgtm
/approve

@terrytangyuan @alculquicondor

roteme-runai:
@rongou, @alculquicondor re-generated since `verify-generate` failed.
There are also failing e2e tests, but they are failing for other, older PRs too. How do we proceed?

alculquicondor:
@tenzen-y has been trying to fix that failure without success.

If you have any free cycles, we would appreciate help on that front. https://github.com/kubeflow/mpi-operator/pull/671

tenzen-y (commented):
Thank you for this contribution. 
Could you rebase on the master branch to take the CI issue fix patch?

PR #697, opened 2025-04-15
Use cncf-hosted gha runners

jeefy:
CNCF has hosted ephemeral GitHub runners in Oracle that we're wanting projects to use rather than the GitHub hosted ones, which are now incur a cost to use. This PR is currently a WIP to work through any tests that break or dependencies that may be missing. <3

Please direct any questions to myself, @krook and @RobertKielty

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,8 +32,7 @@ jobs:
         run: make test
   e2e:
     name: E2E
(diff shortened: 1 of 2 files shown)


tenzen-y:
@jeefy Thank you for working on this. It looks like Kind cluster startup failed.

PR #703, opened 2025-08-14
Enable publishNotReadyAddresses when the runLauncherAsWorker

tenzen-y:
When they enable runLauncherAsWorker, the launcher plays a worker role as well.
In that case, we will face the deadlock between ready for pod vs ready for endpoint in the Launcher. Eventually, MPIJob will fail due to an unready endpoint.

This publishNotReadyAddresses allows us to expose the endpoint immediately, and scripts inside of Launcher can confirm if the Launcher is ready.


terrytangyuan (approved):
Thanks!

/lgtm
/approve

Issue #453, opened 2022-02-10
[Feature request] Add namespace and `svc` suffix for host configmap


shaowei-su:
In the v2 implementation, headless worker service is referenced by `jobname-worker-0.jobname-worker` (https://github.com/alculquicondor/mpi-operator/blob/master/v2/pkg/controller/mpi_job_controller.go#L1110)  which requires the pod `/etc/resolv.conf` config having `namespace.svc.cluster.local` presented.

The `tf-operator` on the other side composes TFCONFIG with both namespace and `svc` suffix: https://github.com/kubeflow/training-operator/blob/master/pkg/controller.v1/tensorflow/tensorflow.go#L159

Is it possible to consolidate the logic so that the same cluster DNS settings work for both operators?

alculquicondor:
Isn't `namespace.svc.cluster.local` already added by kubelet?

In any case, I think it should all continue to work seamlessly if we add `namespace.svc`

shaowei-su:
Not always the case depends on network level configs set up. I validated that adding `namespace.svc` works well like you mentioned.

alculquicondor:
Feel free to send a PR

shaowei-su:
PTAL https://github.com/kubeflow/mpi-operator/pull/454

alculquicondor:
Fixed in #454 
/close

PR #704, opened 2025-08-14
Support cluster domain for MPI HostFile

tenzen-y:
I added `--cluster-domain` option to add the cluster domain when MPIOperator constructs MPI HostFile.
When it is specified the HostFile domains are built with `<pod-name>.<mpi-job-name>.<namespace>.svc.<cluster-domain>` format. Otherwise, `<pod-name>.<mpi-job-name>.<namespace>.svc` is used the same as previously.

#### Background

In https://github.com/kubeflow/mpi-operator/issues/453, we added `<namespace>.svc` suffix to address a special network environment. But, it is not an ideal endpoint name format for the generic network environment, where `svc.<cluster-domain>` is configured by the kubelet. As a result, there are many DNS name resolutions trying until hitting the domain w/out a cluster domain. That causes the Launcher Job to be recreated many times due to the failure of DNS name resolution.

This is quite impacted in the runLauncherAsWorker environment since Launcher is launched very fast before the svc.<cluster-domain> is set up.

### DIFF
--- a/cmd/mpi-operator/app/options/options.go
+++ b/cmd/mpi-operator/app/options/options.go
@@ -40,6 +40,7 @@ type ServerOption struct {
 	Burst               int
 	ControllerRateLimit int
 	ControllerBurst     int
+	ClusterDomain       string
 }
 
 // NewServerOption creates a new CMServer with a default config.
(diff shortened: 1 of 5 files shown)


cmd/mpi-operator/app/options/options.go
```diff
@@ -80,4 +81,8 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 
 	fs.IntVar(&s.ControllerRateLimit, "controller-queue-rate-limit", 10, "Rate limit of the controller events queue .")
 	fs.IntVar(&s.ControllerBurst, "controller-queue-burst", 100, "Maximum burst of the controller events queue.")
+
+	fs.StringVar(&s.ClusterDomain, "cluster-domain", "", `
```
andreyvelich, line 85: @tenzen-y Do we need this for MPI plugin in KF Trainer ?

tenzen-y: I don't think so since Trainer does not support `<PodName>.<JobName>.<namespace>.svc` format. The Trainer supports only `<PodName>.<JobName>`.  In case of Trainer, the cluster domain is automatically considered, immediately.

Issue #691, opened 2025-04-11
Intel OneAPI PGP key does not support new APT requirements


tenzen-y:
# What happens

The `apt update` with the intel OneAPI sourcelist does not succeed since the current Intel OneAPI PGP key format does not satisfy new apt PGP management requirements.

# Background

Since Debian trixie, the apt was switched PGP management tools to `Sequoia` which require more restricted PGP format and does not have backward compatibility compared with older debian apt.

However, the intel OneAPI PGP does not satisfy the new requirement.

# Workaround

We need to unconditionally trust the Intel OneAPI PGP key by the `trusted=yes`option when we register it to the apt sourcelist.

# Appendix
- https://community.intel.com/t5/oneAPI-Registration-Download/intel-oneapi-installation-failes-on-debian-trixie-currently/td-p/1655431
- https://github.com/go-gitea/gitea/issues/33400
- https://wiki.debian.org/OpenPGP/Sequoia
- https://github.com/orgs/community/discussions/27607
- https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg2023150.html#:~:text=After%20apt%20switched%20to%20sqv,fails%20to%20update%2C%20failing%20with

PR #692, opened 2025-04-11
Trust the Intel OneAPI PGP key until it satisfies new APT PGP requirments

tenzen-y:
By these specification changes, all mpi-operator Intel image building failed.

Workaround for https://github.com/kubeflow/mpi-operator/issues/691

PR #706, opened 2025-08-15
Revert "Trust the Intel OneAPI PGP key until it satisfies new APT PGP requirments"

tenzen-y:
Reverts kubeflow/mpi-operator#692

Now, the Debian Trixie has been officially released. So, we can remove a workaround approach.

https://www.debian.org/releases/trixie/

closes: https://github.com/kubeflow/mpi-operator/issues/691


terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #708, opened 2025-08-23
Upgrade K8s version to v1.32.7

tenzen-y:
I upgraded K8s version to v1.32 in Go modules and Test Environment.

As end user benefits, they can leverage the following Features:
- SELinuxChangePolicy (`.spec.mpiReplicaSpecs.*.template.spec.securityContext.seLinuxChangePolicy`): https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#efficient-selinux-volume-relabeling
- PodLevelResources (`.spec.mpiReplicaSpecs.*.template.spec.resources`): https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/


Makefile
```diff
@@ -83,9 +84,11 @@ dev_manifest:
 	sed -e "s~%IMAGE_NAME%~${IMAGE_NAME}~g" -e "s~%IMAGE_TAG%~${RELEASE_VERSION}~g" manifests/overlays/dev/kustomization.yaml.template > manifests/overlays/dev/kustomization.yaml
 
 .PHONY: generate
-generate:
+generate: goimports
 	go generate ./pkg/... ./cmd/...
 	hack/update-codegen.sh
+	# Workaround for https://github.com/kubernetes/kubernetes/issues/129774
+	$(GOIMPORTS) -w pkg/apis/kubeflow/v2beta1/zz_generated.defaults.go
```
tenzen-y, line 91: This is the most big change.
Since code-generator v1.32, the internal implementations for `import` have been changed (`FormatOnly` feature) and are less accurate to improve the generation performance. Thus, unused modules are technically inserted at the head of the Go file (`import` directive) outside of k/k repository.

So, in the short term, if we face such incorrect import directives, we should manually reformat them.

terrytangyuan (approved):
/lgtm
/approve

PR #714, opened 2025-08-31
Upgrade golangci-lint v1 to v2

tenzen-y:
We are using too outdated golangci-lint version. So, it would be better to migrate v1 to v2.
Additionally, I addressed the lint errors since v2 added more and more basic lint rules.


terrytangyuan (approved):
Thanks!

/lgtm
/approve

PR #787, opened 2026-03-15
Fix Running condition being re-emitted when pod and job informers are out-of-sync

GonzaloSaez:
When the pod and job informers are out-of-sync, it's possible for the launcher job to be finished but the pod be running. In that case, the MPIJob may be considered as completed (when using runLauncherAsWorker and workers having finished). In this scenario, the running condition may be re-emitted with a last transition time after the MPIJob was deemed completed. This results in other controllers watching MPIJob to not be able to evaluate the start and end times using the last transition time of the running condition.

To fix this, we can avoid re-emitting the Running condition. Moreover, we can also ensure that the Running condition is always emitted and that the last transition time is <= the completion time.

Another solution to this would be to re-queue if we see the job and pod informers are out-of-sync but I'm not sure if the latter would be harder to implement.

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -1160,6 +1160,30 @@ func (c *MPIJobController) updateMPIJobStatus(mpiJob *kubeflow.MPIJob, launcher 
 	if isMPIJobSuspended(mpiJob) {
 		msg := fmt.Sprintf("MPIJob %s/%s is suspended.", mpiJob.Namespace, mpiJob.Name)
 		updateMPIJobConditions(mpiJob, kubeflow.JobRunning, corev1.ConditionFalse, mpiJobSuspendedReason, msg)
+	} else if isFinished(mpiJob.Status) {
+		// Job reached a terminal state. Do not re-emit Running=True to avoid having a LastTransition time after
+		// the completion time.
+		// If Running was never set (e.g. job completed before the operator observed it running), add Running=False
+		// using the completionTime so that consumers computing duration still find the condition and can make
+		// some guesses about the job lifecycle.
+		if getCondition(mpiJob.Status, kubeflow.JobRunning) == nil {
+			msg := fmt.Sprintf("MPIJob %s/%s is finished but Running condition was never set.", mpiJob.Namespace, mpiJob.Name)
+			cond := kubeflow.JobCondition{
+				Type:    kubeflow.JobRunning,
+				Status:  corev1.ConditionFalse,
+				Reason:  mpiJobRunningReason,
+				Message: msg,
+			}
+			if mpiJob.Status.CompletionTime != nil {
+				cond.LastTransitionTime = *mpiJob.Status.CompletionTime
+				cond.LastUpdateTime = *mpiJob.Status.CompletionTime
+			} else {
+				now := metav1.Now()
+				cond.LastTransitionTime = now
+				cond.LastUpdateTime = now
+			}
+			mpiJob.Status.Conditions = append(mpiJob.Status.Conditions, cond)
+		}
 	} else if launcher != nil && launcherPodsCnt >= 1 && running == len(worker) {
 		msg := fmt.Sprintf("MPIJob %s/%s is running.", mpiJob.Namespace, mpiJob.Name)
 		updateMPIJobConditions(mpiJob, kubeflow.JobRunning, corev1.ConditionTrue, mpiJobRunningReason, msg)
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -1174,14 +1174,10 @@ func (c *MPIJobController) updateMPIJobStatus(mpiJob *kubeflow.MPIJob, launcher 
 				Reason:  mpiJobRunningReason,
 				Message: msg,
 			}
-			if mpiJob.Status.CompletionTime != nil {
-				cond.LastTransitionTime = *mpiJob.Status.CompletionTime
-				cond.LastUpdateTime = *mpiJob.Status.CompletionTime
-			} else {
-				now := metav1.Now()
-				cond.LastTransitionTime = now
-				cond.LastUpdateTime = now
-			}
+			updateTime := ptr.Deref(mpiJob.Status.CompletionTime, metav1.NewTime(c.clock.Now()))
+			cond.LastTransitionTime = updateTime
+			cond.LastUpdateTime = updateTime
+
 			mpiJob.Status.Conditions = append(mpiJob.Status.Conditions, cond)
 		}
 	} else if launcher != nil && launcherPodsCnt >= 1 && running == len(worker) {
--- a/pkg/controller/mpi_job_controller_test.go
+++ b/pkg/controller/mpi_job_controller_test.go
@@ -629,6 +629,71 @@ func TestLauncherSucceeded(t *testing.T) {
 	updateMPIJobConditions(mpiJobCopy, kubeflow.JobCreated, corev1.ConditionTrue, mpiJobCreatedReason, msg)
 	msg = fmt.Sprintf("MPIJob %s/%s successfully completed.", mpiJob.Namespace, mpiJob.Name)
 	updateMPIJobConditions(mpiJobCopy, kubeflow.JobSucceeded, corev1.ConditionTrue, mpiJobSucceededReason, msg)
+	// Running=False is added when the job finishes without Running ever being set.
+	msg = fmt.Sprintf("MPIJob %s/%s is finished but Running condition was never set.", mpiJob.Namespace, mpiJob.Name)
+	updateMPIJobConditions(mpiJobCopy, kubeflow.JobRunning, corev1.ConditionFalse, mpiJobRunningReason, msg)
+	f.expectUpdateMPIJobStatusAction(mpiJobCopy)
+
+	f.run(t.Context(), getKey(mpiJob, t))
+}
+
+// TestLauncherSucceededWithRunningPod tests that when a launcher Job has succeeded but its pod is still observed as Running due to
+// informer lag). Workers have been cleaned up. The Running condition is set to False rather than being re-emitted as True alongside
+// Succeeded.
+func TestLauncherSucceededWithRunningPod(t *testing.T) {
+	f := newFixture(t, "")
+
+	startTime := metav1.Now()
+	completionTime := metav1.Now()
+
+	mpiJob := newMPIJob("test", ptr.To[int32](64), &startTime, &completionTime)
+	// Pre-set the Running condition to simulate a job that was running before completion.
+	msg := fmt.Sprintf("MPIJob %s/%s is created.", mpiJob.Namespace, mpiJob.Name)
+	updateMPIJobConditions(mpiJob, kubeflow.JobCreated, corev1.ConditionTrue, mpiJobCreatedReason, msg)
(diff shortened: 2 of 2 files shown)


tenzen-y:
Sorry for the delayed checking. I found this notification right now...
Let me check this in this week.

pkg/controller/mpi_job_controller.go
```diff
@@ -1160,6 +1160,30 @@ func (c *MPIJobController) updateMPIJobStatus(mpiJob *kubeflow.MPIJob, launcher
 	if isMPIJobSuspended(mpiJob) {
 		msg := fmt.Sprintf("MPIJob %s/%s is suspended.", mpiJob.Namespace, mpiJob.Name)
 		updateMPIJobConditions(mpiJob, kubeflow.JobRunning, corev1.ConditionFalse, mpiJobSuspendedReason, msg)
+	} else if isFinished(mpiJob.Status) {
```
tenzen-y, line 1163:
```suggestion
	} else if isFinished(mpiJob.Status) && getCondition(mpiJob.Status, kubeflow.JobRunning) == nil{
```

It seems that we can simplify the if-else structure by doing this.

pkg/controller/mpi_job_controller.go
```diff
@@ -1160,6 +1160,30 @@ func (c *MPIJobController) updateMPIJobStatus(mpiJob *kubeflow.MPIJob, launcher
...
+			if mpiJob.Status.CompletionTime != nil {
+				cond.LastTransitionTime = *mpiJob.Status.CompletionTime
+				cond.LastUpdateTime = *mpiJob.Status.CompletionTime
+			} else {
+				now := metav1.Now()
+				cond.LastTransitionTime = now
+				cond.LastUpdateTime = now
+			}
+			mpiJob.Status.Conditions = append(mpiJob.Status.Conditions, cond)
+		}
```
tenzen-y, line 1186:
```suggestion
		msg := fmt.Sprintf("MPIJob %s/%s is finished but Running condition was never set.", mpiJob.Namespace, mpiJob.Name)
		cond := kubeflow.JobCondition{
			Type:    kubeflow.JobRunning,
			Status:  corev1.ConditionFalse,
			Reason:  mpiJobRunningReason,
			Message: msg,
		}
		updateTime := ptr.Deref(mpiJob.Status.CompletionTime, c.clock.Now())
		cond.LastTransitionTime := updateTime
		cond.LastUpdateTime := updateTime
		mpiJob.Status.Conditions = append(mpiJob.Status.Conditions, cond)
```

1. Unnest the condition
2. Use the clock
3. Simplify pointer deref operation

pkg/controller/mpi_job_controller_test.go
```diff
@@ -629,6 +629,71 @@ func TestLauncherSucceeded(t *testing.T) {
...
+}
+
+// TestLauncherSucceededWithRunningPod tests that when a launcher Job has succeeded but its pod is still observed as Running due to
+// informer lag). Workers have been cleaned up. The Running condition is set to False rather than being re-emitted as True alongside
+// Succeeded.
+func TestLauncherSucceededWithRunningPod(t *testing.T) {
+	f := newFixture(t, "")
+
+	startTime := metav1.Now()
+	completionTime := metav1.Now()
```
tenzen-y, line 647: Could you use `clocktesting`?

pkg/controller/mpi_job_controller_test.go
```diff
@@ -629,6 +629,71 @@ func TestLauncherSucceeded(t *testing.T) {
...
+	updateMPIJobConditions(mpiJobCopy, kubeflow.JobRunning, corev1.ConditionFalse, mpiJobRunningReason, msg)
+	f.expectUpdateMPIJobStatusAction(mpiJobCopy)
+
+	f.run(t.Context(), getKey(mpiJob, t))
+}
+
+// TestLauncherSucceededWithRunningPod tests that when a launcher Job has succeeded but its pod is still observed as Running due to
+// informer lag). Workers have been cleaned up. The Running condition is set to False rather than being re-emitted as True alongside
+// Succeeded.
+func TestLauncherSucceededWithRunningPod(t *testing.T) {
```
tenzen-y, line 643: Can you explicitly check if the Running LastConditions are later or equal to completionTime?

Because testing libraries ignore such time comparison: https://github.com/GonzaloSaez/mpi-operator/blob/8c953844bc5a5deb7ba98d5e1e85a324b6aa2645/pkg/controller/mpi_job_controller_test.go#L350

tenzen-y (commented):
Thank you for fixing that!
I left some comments.

tenzen-y:
@GonzaloSaez Friendly ping

pkg/controller/mpi_job_controller.go
GonzaloSaez: We cannot do that in this case because then we can enter the next else if condition. We have to nest unfortunately

GonzaloSaez:
@tenzen-y sorry for taking so long to reply. I think I have addressed your comments

pkg/controller/mpi_job_controller.go
tenzen-y: Ah, you're right. Thank you for the confirmation.

pkg/controller/mpi_job_controller_test.go
```diff
@@ -369,6 +373,27 @@ func checkAction(expected, actual core.Action, t *testing.T) {
 	}
 }
 
+func assertRunningConditionNotBeforeCompletionTime(t *testing.T, mpiJob *kubeflow.MPIJob) {
+	t.Helper()
+
+	if mpiJob == nil || mpiJob.Status.CompletionTime == nil {
```
tenzen-y:
```suggestion
	if mpiJob == nil || mpiJob.Status.CompletionTime == nil || !isFinished(mpiJob.Status) {
```

Shouldn't we exclude unfinished MPIJob?

pkg/controller/mpi_job_controller_test.go
```diff
@@ -369,6 +373,27 @@ func checkAction(expected, actual core.Action, t *testing.T) {
...
+
+	running := getCondition(mpiJob.Status, kubeflow.JobRunning)
+	if running == nil {
+		return
+	}
+
+	completionTime := mpiJob.Status.CompletionTime.Time
+	if running.LastTransitionTime.Time.Before(completionTime) {
+		t.Errorf("MPIJob %s/%s Running LastTransitionTime %s is before CompletionTime %s", mpiJob.Namespace, mpiJob.Name, running.LastTransitionTime.Time, completionTime)
+	}
```
tenzen-y:
```suggestion
	if completionTime.Before(running.LastTransitionTime.Time) {
		t.Errorf("... Running LastTransitionTime %s is after CompletionTime %s", ..., running.LastTransitionTime.Time, completionTime)
	}
```

Isn't this inverted?

pkg/controller/mpi_job_controller_test.go
```diff
@@ -369,6 +373,27 @@ func checkAction(expected, actual core.Action, t *testing.T) {
...
+	if running.LastUpdateTime.Time.Before(completionTime) {
+		t.Errorf("MPIJob %s/%s Running LastUpdateTime %s is before CompletionTime %s", mpiJob.Namespace, mpiJob.Name, running.LastUpdateTime.Time, completionTime)
+	}
```
tenzen-y:
```suggestion
	if completionTime.Before(running.LastUpdateTime.Time) {
		t.Errorf("... Running LastUpdateTime %s is after CompletionTime %s", ..., running.LastUpdateTime.Time, completionTime)
	}
```

ditto

pkg/controller/mpi_job_controller_test.go
```diff
@@ -629,6 +654,72 @@ func TestLauncherSucceeded(t *testing.T) {
...
+	f.expectUpdateMPIJobStatusAction(mpiJobCopy)
+
+	f.run(t.Context(), getKey(mpiJob, t))
+}
+
+// TestLauncherSucceededWithRunningPod tests that when a launcher Job has succeeded but its pod is still observed as Running due to
+// informer lag). Workers have been cleaned up. The Running condition is set to False rather than being re-emitted as True alongside
+// Succeeded.
+func TestLauncherSucceededWithRunningPod(t *testing.T) {
+	fakeClock := clocktesting.NewFakeClock(time.Now().Truncate(time.Second))
```
tenzen-y:
```suggestion
	fakeClock := clocktesting.NewFakeClock(time.Now().Add(-time.Hour).Truncate(time.Second))
```

Add - 1 hour to make the fake clock time in the past.

pkg/controller/mpi_job_controller_test.go
```diff
@@ -629,6 +654,72 @@ func TestLauncherSucceeded(t *testing.T) {
...
+	startTime := metav1.NewTime(fakeClock.Now())
+	completionTime := metav1.NewTime(fakeClock.Now())
+
+	mpiJob := newMPIJob("test", ptr.To[int32](64), &startTime, &completionTime)
+	// Pre-set the Running condition to simulate a job that was running before completion.
+	msg := fmt.Sprintf("MPIJob %s/%s is created.", mpiJob.Namespace, mpiJob.Name)
+	updateMPIJobConditions(mpiJob, kubeflow.JobCreated, corev1.ConditionTrue, mpiJobCreatedReason, msg)
+	msg = fmt.Sprintf("MPIJob %s/%s is running.", mpiJob.Namespace, mpiJob.Name)
+	updateMPIJobConditions(mpiJob, kubeflow.JobRunning, corev1.ConditionTrue, mpiJobRunningReason, msg)
+	f.setUpMPIJob(mpiJob)
```
tenzen-y, line 688:
```suggestion
    for i := range mpiJob.Status.Conditions {
            if mpiJob.Status.Conditions[i].Type == kubeflow.JobRunning {
                    mpiJob.Status.Conditions[i].LastTransitionTime = startTime
                    mpiJob.Status.Conditions[i].LastUpdateTime = startTime
            }
    }
	f.setUpMPIJob(mpiJob)
```

emulate time.

pkg/controller/mpi_job_controller_test.go
```diff
@@ -629,6 +654,72 @@ func TestLauncherSucceeded(t *testing.T) {
...
+
+// TestLauncherSucceededWithRunningPod tests that when a launcher Job has succeeded but its pod is still observed as Running due to
+// informer lag). Workers have been cleaned up. The Running condition is set to False rather than being re-emitted as True alongside
+// Succeeded.
+func TestLauncherSucceededWithRunningPod(t *testing.T) {
+	fakeClock := clocktesting.NewFakeClock(time.Now().Truncate(time.Second))
+	f := newFixture(t, "")
+
+	startTime := metav1.NewTime(fakeClock.Now())
+	completionTime := metav1.NewTime(fakeClock.Now())
```
tenzen-y, line 674:
```suggestion
	fakeClock.Step(time.Minute)
	completionTime := metav1.NewTime(fakeClock.Now())
```

pkg/controller/mpi_job_controller_test.go
```diff
@@ -369,6 +373,27 @@ func checkAction(expected, actual core.Action, t *testing.T) {
 	}
 }
 
+func assertRunningConditionNotBeforeCompletionTime(t *testing.T, mpiJob *kubeflow.MPIJob) {
```
tenzen-y:
```suggestion
func assertRunningConditionNotAfterCompletionTime(t *testing.T, mpiJob *kubeflow.MPIJob) {
```

tenzen-y (commented):
Thank you!
/lgtm
/approve

PR #810, opened 2026-06-10
Prevent terminating launcher pods from being recreated

GonzaloLuminary:
Avoid running into https://github.com/kubernetes/kubernetes/issues/115844 where pods can be recreated while the launcher pod is terminating resulting in MPIJob duplicating work when using runLauncherAsWorker

### DIFF
--- a/pkg/controller/mpi_job_controller.go
+++ b/pkg/controller/mpi_job_controller.go
@@ -1542,6 +1542,14 @@ func (c *MPIJobController) newLauncherJob(mpiJob *kubeflow.MPIJob) *batchv1.Job 
 			ActiveDeadlineSeconds:   mpiJob.Spec.RunPolicy.ActiveDeadlineSeconds,
 			BackoffLimit:            mpiJob.Spec.RunPolicy.BackoffLimit,
 			Template:                c.newLauncherPodTemplate(mpiJob),
+			// Make sure we don't run into https://github.com/kubernetes/kubernetes/issues/115844 where
+			// terminating pods are recreated under certain conditions.
+			// This was initially fixed in https://github.com/kubernetes/kubernetes/pull/117015 which landed
+			// in 1.28. PodReplacementPolicy became GA in k8s 1.34 and the check for PodFailurePolicy was
+			// removed in the middle. Whether the workarounds work depend on the k8s version and the
+			// feature flags being active but it's the best we can do.
+			PodReplacementPolicy: ptr.To(batchv1.Failed),
+			PodFailurePolicy:     &batchv1.PodFailurePolicy{},
 		},
 	}
 	if isMPIJobSuspended(mpiJob) {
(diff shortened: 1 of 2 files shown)


GonzaloLuminary:
cc: @tenzen-y

pkg/controller/mpi_job_controller.go
```diff
@@ -1542,6 +1542,14 @@ func (c *MPIJobController) newLauncherJob(mpiJob *kubeflow.MPIJob) *batchv1.Job
...
 			BackoffLimit:            mpiJob.Spec.RunPolicy.BackoffLimit,
 			Template:                c.newLauncherPodTemplate(mpiJob),
+			// Make sure we don't run into https://github.com/kubernetes/kubernetes/issues/115844 where
+			// terminating pods are recreated under certain conditions.
+			// This was initially fixed in https://github.com/kubernetes/kubernetes/pull/117015 which landed
+			// in 1.28. PodReplacementPolicy became GA in k8s 1.34 and the check for PodFailurePolicy was
+			// removed in the middle. Whether the workarounds work depend on the k8s version and the
+			// feature flags being active but it's the best we can do.
+			PodReplacementPolicy: ptr.To(batchv1.Failed),
+			PodFailurePolicy:     &batchv1.PodFailurePolicy{},
```
tenzen-y: I agree with introducing the `podReplacementPolicy: "Failed"` motivation, but why do we need to specify an empty PodFailurePolicy? I don't see any reason.


```suggestion
```

tenzen-y (commented):
@GonzaloLuminary Thank you for working on this PR!
Can you also address CI failures?

pkg/controller/mpi_job_controller.go
GonzaloLuminary: Depending on the k8s version and the feature flags being active, setting the PodReplacementPolicy may not be enough. As per the original PR that addressed the issue https://github.com/kubernetes/kubernetes/pull/117015, the condition to activate the fix was controlled by this function

```
func onlyReplaceFailedPods(job *batch.Job) bool {
	if feature.DefaultFeatureGate.Enabled(features.JobPodReplacementPolicy) && *job.Spec.PodReplacementPolicy == batch.Failed {
		return true
	}
	return feature.DefaultFeatureGate.Enabled(features.JobPodFailurePolicy) && job.Spec.PodFailurePolicy != nil
}
```

In master the same function reads

```
func onlyReplaceFailedPods(job *batch.Job) bool {
	return job.Spec.PodReplacementPolicy != nil && *job.Spec.PodReplacementPolicy == batch.Failed
}
```

It all depends which k8s versions/configurations we want to support. I'm happy to drop support for old clusters with JobPodReplacementPolicy not being active.

tenzen-y: AFAIK, the JobPodReplacementPolicy is enabled by default since K8s 1.29: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#:~:text=1.28-,JobPodReplacementPolicy,-true And K8s 1.28 is no longer supported in upstream K8s and almost cloud providers. So, I think that we should not care such deprecated K8s versions.

tenzen-y (commented):
Thank you!
/lgtm
/approve

PR #819, opened 2026-06-30
Fix missing DeepCopy when mutating the launcher job

GonzaloLuminary:
We were mutating the informer's cache when updating the launcher Job. If the update failed, the cache would be poisoned, leaving later reconciles to operate on a launcher description that doesn't match the real state on the API server.


GonzaloLuminary:
cc: @tenzen-y please take a look at this issue whenever you have time. I think it is an important bugfix

tenzen-y:
> cc: @tenzen-y please take a look at this issue whenever you have time. I think it is an important bugfix

Thank you for raising this PR. I am waiting for CI checks. After it succeeded, I will merge this PR.

tenzen-y (commented):
All green.

/lgtm
/approve

Issue #596, opened 2023-10-12
Update stale examples


jarulsamy:
Hi, 

Some of the examples are no longer working. For example, [pi-mpich.yaml](https://github.com/kubeflow/mpi-operator/blob/2747ca701bcf5a3a68ca92269e1c31ebfde40149/examples/v2beta1/pi/pi-mpich.yaml) refers to docker image tags that no longer exist. In addition, it would be really nice if we had some larger examples (similar to the tensorflow ones) that don't need a GPU. 

If there are no objections, I am happy to get a PR put together for this myself.

terrytangyuan:
Feel free to submit PRs!

tenzen-y:
Regarding pi-mpich.yaml, the image isn't outdated. Although we release the example images every release, the image hasn't yet been released since the MPICH implementation was contributed recently. This means that we don't have any releases for the MPICH implementation.

PR #822, opened 2026-07-01
fix: correct broken README example paths and kustomize command

immanuwell:
This fixes a couple README commands that break on a fresh clone. Small papercut, but pretty user-facing.

Bug
`cat examples/pi/pi-intel.yaml`
`cat examples/pi/pi-mpich.yaml`
both fail with `No such file or directory`

`kubectl kustomize base | kubectl apply -f -`
also points to a `base/` dir that is not in this repo.

Fix
Use `examples/v2beta1/pi/...` and `manifests/overlays/kubeflow` instead.

Repro
Run those commands from repo root.

Related #596

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,16 +48,15 @@ cd mpi-operator
 kustomize build manifests/overlays/kubeflow | kubectl apply -f -
 ```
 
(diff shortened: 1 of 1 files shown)


tenzen-y (commented):
Thank you!
/lgtm
/approve

PR #827, opened 2026-07-04
fix: correct cleanPodPolicy default in generated docs

immanuwell:
Problem

`SetDefaults_MPIJob` defaults omitted `cleanPodPolicy` to `None`, but the shipped CRD, deploy yaml, swagger, and python sdk docs say `Running`.
That is a pretty easy footgun in the wild.

Fix

Update the source comment and regenerate the published API surfaces so they all say `Defaults to None`.

Repro

1. Open `pkg/apis/kubeflow/v2beta1/default.go` and see `policy.CleanPodPolicy = ptr.To(CleanPodPolicyNone)`.
2. Open the pre-fix generated docs, for example `deploy/v2beta1/mpi-operator.yaml`, and see `Default to Running`.
3. Omit `spec.runPolicy.cleanPodPolicy` in an `MPIJob`. The real default is `None`, so the docs were off.

Validation

`go test ./pkg/apis/kubeflow/v2beta1 ./pkg/apis/kubeflow/validation ./pkg/controller`


tenzen-y:
@immanuwell Thank you for your contribution, it seems that generated codes are out dated. Can you verify and fix that?

https://github.com/kubeflow/mpi-operator/actions/runs/28703199713/job/85237964347?pr=827

immanuwell:
@tenzen-y done, regenerated it, PTAL

tenzen-y (commented):
Thank you!
/lgtm
/approve

PR #830, opened 2026-07-07
Fix missing rbac for job status update

GonzaloLuminary:
Without this, we cannot update the job status when unsuspending an MPIJob and the launcher job has a StartTime set.

### DIFF
--- a/manifests/base/cluster-role.yaml
+++ b/manifests/base/cluster-role.yaml
@@ -64,8 +64,6 @@ rules:
   resources:
   - jobs/status
   verbs:
-  - get
-  - patch
   - update
 - apiGroups:
   - apiextensions.k8s.io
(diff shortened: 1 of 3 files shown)


tenzen-y (commented):
Thank you 👍 
/lgtm
/approve

manifests/base/cluster-role.yaml
```diff
@@ -59,6 +59,14 @@ rules:
...
   - update
   - watch
+- apiGroups:
+  - batch
+  resources:
+  - jobs/status
+  verbs:
+  - get
+  - patch
+  - update
```
GonzaloLuminary: I think we only need update in here, not sure if it's worth rerunning the suite

tenzen-y: Let's see if only update is needed or not

tenzen-y: Seems perfect

/lgtm

PR #738, opened 2025-10-29
Implement clusterDomain propagation to discover file

tenzen-y:
The same as https://github.com/kubeflow/mpi-operator/pull/704, the cluster domain supports discover file (`discover_hosts.sh`) as well.

PR #832, opened 2026-07-10
fix: honor cluster-domain in worker DNS search paths

immanuwell:
small follow-up to #704 and #738.

`--cluster-domain` already updates the hostfile and `discover_hosts.sh`, but worker pods still get `dnsConfig.searches` hardcoded to `svc.cluster.local`.
That bites Intel MPI and MPICH, workers need to resolve the launcher hostname too.

Fix:
- use the configured cluster domain when building worker DNS search suffixes
- add a regression test for a custom-domain Intel MPI job

Repro:
1. Start the operator with `--cluster-domain=kube.local`
2. Create an Intel MPI or MPICH job
3. Inspect a worker pod with `kubectl get pod <worker> -o yaml`
4. Before this patch it has `dnsConfig.searches: ["<job>.<ns>.svc.cluster.local"]`
5. The launcher FQDN is `...svc.kube.local`, so hostname lookup is off and the job can break


tenzen-y (commented):
Ah, this is actually bug for MPICH and Intel MPI in case of the custom domain.
Thank you!

/lgtm
/approve

PR #833, opened 2026-07-11
fix: allow zero workers when launcher acts as worker

immanuwell:
small fix for a sneaky `runLauncherAsWorker` footgun.

if `runLauncherAsWorker: true` is used and the `Worker` spec omits `replicas`, defaulting sets worker replicas to `0`. reconcile then validates the defaulted job, hits `ValidationError`, and bails out, so the launcher Job never shows up.

this patch allows `Worker.replicas: 0` only for that mode. other zero worker cases still fail, so the old guardrail stays.

repro:
1. apply this:
```yaml
apiVersion: kubeflow.org/v2beta1
kind: MPIJob
metadata:
  name: repro
spec:
  runLauncherAsWorker: true
  runPolicy:
    cleanPodPolicy: Running
  mpiReplicaSpecs:
    Launcher:
      template:
        spec:
          containers:
          - name: main
            image: mpi-image
    Worker:
      template:
        spec:
          containers:
          - name: main
            image: mpi-image
```
2. on current `master`, the job gets a `ValidationError` event and no launcher Job is created
3. with this patch, the launcher Job, Service, ConfigMap and Secret are created, and there are no worker Pods, which is the intended path

tests:
- validation coverage for zero workers in launcher-as-worker mode
- controller reconcile regression test
- integration regression test

### DIFF
--- a/pkg/apis/kubeflow/validation/validation.go
+++ b/pkg/apis/kubeflow/validation/validation.go
@@ -68,7 +68,7 @@ func validateMPIJobName(job *kubeflow.MPIJob) field.ErrorList {
 }
 
 func validateMPIJobSpec(spec *kubeflow.MPIJobSpec, path *field.Path) field.ErrorList {
-	errs := validateMPIReplicaSpecs(spec.MPIReplicaSpecs, path.Child("mpiReplicaSpecs"))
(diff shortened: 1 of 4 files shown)


tenzen-y:
@immanuwell Could you sign the commit for the DCO check?

PR #869, opened 2026-09-22
Enforce the allowed values of launcherCreationPolicy

poojarishreyas:
## What this PR does / why we need it

The Enum marker on `MPIJobSpec.LauncherCreationPolicy` is written as:

```go
// +kubebuilder:validation:Enum:AtStartup;WaitForWorkersReady
// +kubebuilder:default:=AtStartup
LauncherCreationPolicy LauncherCreationPolicy `json:"launcherCreationPolicy,omitempty"`
```

`controller-gen` only recognizes the `Enum:=` / `Enum=` spelling, so this marker is silently ignored and **no `enum` is generated in the CRD** — unlike `mpiImplementation`, declared a few lines below with `Enum:=`, which does get one:

```yaml
              launcherCreationPolicy:
                default: AtStartup
                description: launcherCreationPolicy if WaitForWorkersReady, ...
                type: string          # <- no enum
              mpiImplementation:
                default: OpenMPI
                description: ...
                enum:
                - OpenMPI
                - Intel
                - MPICH
                type: string
```

So the API server accepts any string for this field. The controller only compares the value against `AtStartup`, which means a typo such as `atStartup` silently selects the `WaitForWorkersReady` behavior: the launcher is delayed until all workers are ready, with no feedback to the user.

This PR:

1. fixes the marker and regenerates `manifests/base` and `deploy/v2beta1` (the generated diff is exactly the 3 added `enum` lines, in both files);
2. validates the field in `ValidateMPIJob`, the way `mpiImplementation`, `cleanPodPolicy`, `restartPolicy` and `managedBy` already are — so that an MPIJob created against a CRD that predates this fix surfaces a `ValidationError` event instead of running under an unintended policy.

Because the field is defaulted (by the CRD and by `SetDefaults_MPIJob`), the validation treats the empty value as unsupported, consistent with `mpiImplementation`; the existing test cases were updated accordingly.

## Checklist

- [x] Manifests regenerated with `controller-gen` v0.20.1 and `kustomize` v5.7.1 (the pinned versions); regenerating before the marker fix reproduces the committed CRD byte-for-byte, so the diff contains nothing but the enum.
- [x] `pkg/apis/kubeflow/validation` tests updated and passing; `go test ./pkg/...` passes.


pkg/apis/kubeflow/validation/validation.go
tenzen-y, line 1: We don't need this change since we can validate via CRD schema validations.

tenzen-y (commented):
Thank you for sending this PR.
I left a few comments.

pkg/apis/kubeflow/validation/validation.go
poojarishreyas: Thanks for the review!

Agreed — dropped the validation change, the CRD schema covers it. The PR is now just the marker fix plus the regenerated manifests.

PTAL

